PR #1, opened 2023-02-08
feat: switch to typescript

laurentsenta:
- Move all the code to typescript (fixture generation and tests),
- Future-proofed with esm modules.
- Not touching the implementation other than adding a few checks to conform to types. I expect we'll revisit all the code in the next couple of weeks.


galargh (approved):
🚀

PR #4, opened 2023-02-14
fixture: generate fixtures + use a car file export

laurentsenta:


laurentsenta:
(closing we're moving to a go stack)

PR #5, opened 2023-03-04
refactor: simplify action.yml, add docker-container-action for testin…

galargh:
…g and fixture extraction

This commit simplifies the action.yml for the extract-fixtures and test actions, removing the need for a separate docker build step. Additionally, it uses the newly added docker-container-action for both of these actions.


galargh:
Docker Container Action: https://github.com/pl-strflt/docker-container-action

PR #6, opened 2023-03-06
Rewrite in Go

galargh:


laurentsenta (approved):
Now that we validate the approach with @lidel, Let's merge as is and improve from there.

galargh:
> @laurentsenta will the nicer GO-diomatic notation be applied in follow-up PR? Can we have an issue with example of this nicer notation?

Yes! We're gonna rewrite existing tests in the nice notation. Let's just merge it as-is as a base so that we can start working using PRs.

tests/t0118_gateway_car_test.go
```diff
@@ -0,0 +1,79 @@
...
+					"Accept-Ranges": IsEqualWithHint(
+						"CAR is streamed, gateway may not have the entire thing, unable to support range-requests. Partial downloads and resumes should be handled using IPLD selectors: https://github.com/ipfs/go-ipfs/issues/8769",
+						"none",
+					),
+				},
+			},
+		},
+		{
+			// Test between l85 and l112
+			Name: "GET response for application/vnd.ipld.car2",
```
galargh, line 50: @lidel It's already used here for example.

PR #9, opened 2023-03-07
feat: make subdomain testing possible

laurentsenta:
This PR adds support for subdomain testing. We port the test from sharness 114 into the conformance test suite.

- the ability to use any URL (like dweb gateway).
- nicer reporting, see reports in https://github.com/ipfs/gateway-conformance/actions/runs/4470741917/jobs/7854699679

## Tasks

- [x] Update the CI's ipfs daemon with gateway URLs (localhost and example.com should be enabled)
- [x] port missing test (50%)
- [x] resolve a few discussions tagged for lidel
- [x] implement TODOs
  - [x] get the correct value from CID


tests/t0114_gateway_subdomains_test.go
```diff
@@ -11,46 +11,32 @@ import (
...
-	// // CIDv0to1 is necessary because raw-leaves are enabled by default during
-	// // "ipfs add" with CIDv1 and disabled with CIDv0
-	// CIDv0to1 := "bafybeiffndsajwhk3lwjewwdxqntmjm4b5wxaaanokonsggenkbw6slwk4"
-	// CIDv1_TOO_LONG := "bafkrgqhhyivzstcz3hhswshfjgy6ertgmnqeleynhwt4dlfsthi4hn7zgh4uvlsb5xncykzapi3ocd4lzogukir6ksdy6wzrnz6ohnv4aglcs"
-	DIR_CID := "bafybeiht6dtwk3les7vqm6ibpvz6qpohidvlshsfyr7l5mpysdw2vmbbhe" // ./testdirlisting
+	DIR_CID := "bafybeiht6dtwk3les7vqm6ibpvz6qpohidvlshsfyr7l5mpysdw2vmbbhe"
 
 	tests := []CTest{
 		{
-			Name: "request for 127.0.0.1/ipfs/{CID} stays on path",
```
laurentsenta, line 28: To be checked w/Lidel, I guess that test is kubo-centric.

tests/t0114_gateway_subdomains_test.go
```diff
@@ -13,44 +13,28 @@ import (
...
-				Body:       Contains(CID_VAL),
-			},
-		},
-		{
-			Name: "request for localhost/ipfs/{CIDv1} returns HTTP 301 Moved Permanently",
+			Name: "request for {gateway}/ipfs/{CIDv1} returns HTTP 301 Moved Permanently",
 			Request: CRequest{
 				DoNotFollowRedirects: true,
-				Url:                  fmt.Sprintf("/ipfs/%s", CIDv1),
+				RawURL: fmt.Sprintf("http://%s/ipfs/%s", SubdomainGatewayHost, CIDv1),
```
laurentsenta, line 24: To discuss, that syntax is not great.

laurentsenta:
cc @galargh this is the subdomain with custom resolver trick.
Behind a flag for now, it requires the user to call beforehand.

```
ipfs config --json Gateway.PublicGateways '{
  "example.com": {
    "UseSubdomains": true,
    "Paths": ["/ipfs", "/ipns", "/api"]
  }
}'
```

tests/t0114_gateway_subdomains_test.go
```diff
@@ -13,49 +13,63 @@ import (
...
+		{
+			Name: "request for {cid}.ipfs.localhost/api returns data if present on the content root (sugar)",
+			Request: Request().
+				URL("http://%s.ipfs.example.com/api/file.txt", DIR_CID).
+				Request(),
+			Response: Expect().
+				Status(200).
+				Body("I am a txt file\n").
+				Response(),
+		},
```
laurentsenta, line 72: Note the test above, without sugar, we have to use URL:
`Url: fmt.Sprintf("%s://%s.ipfs.%s/api/file.txt", SubdomainGatewayScheme, DIR_CID, SubdomainGatewayHost),`

the sugar rewrites the URLs for you, which means a test implementer can use
`URL("http://%s.ipfs.example.com/api/file.txt", DIR_CID).`
and the testing tool rewrites it.

tests/t0114_gateway_subdomains_test.go
```diff
@@ -13,49 +13,63 @@ import (
...
+			Response: Expect().
+				Status(301).
+				Headers(
+					Header("Location").
+						// TODO: this works only because we use example.com in our tests.
+						// It should be:
+						// Contains("%s://%s.ipfs.%s", SubdomainGatewayScheme, CIDv1, SubdomainGatewayHost)
+						// I am trying to avoid this syntax.
+						// The other option is to force the tested gateway to use example.com.
+						Contains("http://%s.ipfs.example.com", CIDv1),
```
laurentsenta, line 50: I don't have a good solution for now, apart from wrapping with some function that signals `ThisIsASubdomainURL`.

tests/t0114_gateway_subdomains_test.go
```diff
@@ -2,61 +2,562 @@ package tests
...
+		// # ============================================================================
+
+		// # <libp2p-key>.ipns.example.com
+
+		// # API on subdomain gateway example.com
+		// # ============================================================================
+
+		// # disable /api on example.com
+		// TODO(lidel): in this part of the test suite, we disable example.com then
+		// we test that the endpoints returns 404. This is more Kubo than conformance-related.
```
laurentsenta, line 373: The test is in: 
https://github.com/ipfs/kubo/blob/668d0b2fa4b43a1474ea0d1deab5295810a55f68/test/sharness/t0114-gateway-subdomains.sh#L537

My guess is that this should not live in the conformance test suite.

tests/t0114_gateway_subdomains_test.go
```diff
@@ -2,61 +2,562 @@ package tests
...
+			Name: fmt.Sprintf("%s (HTTP proxy)", label),
+			Hint: fmt.Sprintf("%s\n%s", hint, "HTTP proxy (hostname is passed via URL)"),
+			Request: baseReq.
+				URL(baseUrl).
+				Proxy(GatewayUrl).
+				DoNotFollowRedirects().
+				Request(),
+			Response: expected,
+			// TODO(lidel): do we need to test different http behavior ?
+			// The previous test was using 2 proxy (--proxy and --proxy1.0)
```
laurentsenta, line 542: @lidel below, in [proxy.go](https://github.com/ipfs/gateway-conformance/pull/9/files#diff-df03cc864581cbd298d0dce0186ff5350c18c5733c46ccaaa32b929d38ab673eR1) we implement the proxy and tunnel proxy equivalents. The test used proxy1.0 before; ([ref](https://github.com/ipfs/kubo/blob/668d0b2fa4b43a1474ea0d1deab5295810a55f68/test/sharness/t0114-gateway-subdomains.sh#L56)) should we support HTTP 1.0, 1.1, and 2.0 explicitly for that first iteration?

tests/t0114_gateway_subdomains_test.go
```diff
@@ -2,61 +2,562 @@ package tests
...
+
+		// # example.com/ip(f|n)s/*
+		// # =============================================================================
+
+		// # path requests to the root hostname should redirect
+		// # to a subdomain URL with proper origin isolation
+
+		// TODO(lidel): for some reason in 114, we only test the simple case, with Host, no proxy or tunnel.
+		// Do we need to port `test_hostname_gateway_response_should_contain` or can we
+		// reuse the same "larger" test?
```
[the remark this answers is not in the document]
lidel: I think this could reuse part of the larger test, if we can test localhost and public dns name with the same code, that is better.

I don't remember the reason these have a separate function, but suspect it was to avoid dealing with `--resolve $hostname:127.0.0.1` so we could run tests against a public gateway, but that never happened.

tests/t0114_gateway_subdomains_test.go
lidel: We only care about real world proxy modes, afaik only ones that matter are the three supported by `curl` (default vs HTTP proxy CONNECT 1.1 or 1.0)

- https://curl.se/docs/manpage.html#-x
- https://curl.se/docs/manpage.html#--proxy10  
- https://curl.se/docs/manpage.html#-p


I did not see anything specific to HTTP/2 in the wild, we don't need to worry about it yet.

laurentsenta: Thanks for the details! I looked into the docs and implementations,

**We don't test HTTP 1.0; there's a typo in sharness.**

Note the doc for proxytunnel:

> The only difference between this and the HTTP proxy option [-x, --proxy](https://curl.se/docs/manpage.html#-x), is that attempts to use CONNECT through the proxy will specify an HTTP 1.0 protocol instead of the default HTTP 1.1.

We don't use `--proxytunnel` with `--proxy1.0` which means that our tests [A](https://github.com/ipfs/kubo/blob/668d0b2fa4b43a1474ea0d1deab5295810a55f68/test/sharness/t0114-gateway-subdomains.sh#L56) and [B](https://github.com/ipfs/kubo/blob/668d0b2fa4b43a1474ea0d1deab5295810a55f68/test/sharness/t0114-gateway-subdomains.sh#L49) are testing the same thing.

Demo gist with every outputs: https://gist.github.com/laurentsenta/18b37a8405e4fee0d819102dde52e95b

**golang seems to ignore HTTP 1.0**

https://pkg.go.dev/net/http#Request

```
// For client requests, these fields are ignored. The HTTP
// client code always uses either HTTP/1.1 or HTTP/2.
// See the docs on Transport for details.
```

The test would be non-trivial to implement

**But**

We're not really exercising HTTP 1.0 anyway, we never keep the connection between requests, and we always pass a Host header. At best, we'd check that the gateway accepts our user agent.

My recommendation:
- we create an issue to add test HTTP 1.0 eventually but don't act on it for now.

lidel: Nice catch! Sgtm. If nobody complained about lack of HTTP 1.0 here, means we can skip / remove.

tooling/test/sugar.go
```diff
@@ -2,58 +2,143 @@ package test
...
+	"strings"
 
 	"github.com/ipfs/gateway-conformance/tooling/check"
 )
 
+// When a test writer uses a URL that contains the example.com domain, we assume they
+// mean "the current subdomain URL". This means that "http://something.ipfs.example.com/path"
+// is rewritten to hit the current tested gateway, which might live under a different name, and
+// different port.
+func TestURLFromSpecURL(testURL string) string {
```
galargh, line 15: We're not using this, right?

tooling/test/sugar.go
```diff
@@ -2,58 +2,143 @@ package test
...
-		r = r.Header(h)
+		r.Headers_[h.Key] = h.Value
 	}
 
 	return r
 }
 
 func (r RequestBuilder) Request() CRequest {
+	if r.URL_ != "" && r.Path_ != "" {
+		panic("Both 'Url' and 'Path' are set")
```
galargh, line 119:
```suggestion
		panic("Both 'URL' and 'Path' are set")
```

tooling/test/sugar.go
```diff
@@ -2,58 +2,143 @@ package test
...
+	}
+
 	return CRequest{
-		Method:  r.Method_,
-		Url:     r.Url_,
-		Headers: r.Headers_,
+		Method:               r.Method_,
+		Path:                 r.Path_,
+		Url:                  r.URL_,
```
galargh, line 125: Could we make this `URL` too for consistency?

galargh (approved):
Nicely done @laurentsenta 🏅 👏 

I left some comments, but I don't want to block this. It might be worth agreeing on the param name before the merge, but even that isn't something that we can't tackle later.

tooling/test/sugar.go
laurentsenta: forgotten dead code, thanks for catching it

tests/t0114_gateway_subdomains_test.go
```diff
@@ -2,61 +2,471 @@ package tests
...
+			Expect().
+				Status(301).
+				Headers(
+					Header("Location").
+						Hint("request for example.com/ipfs/{CIDv1} returns Location HTTP header for subdomain redirect in browsers").
+						Contains("%s://%s.ipfs.%s/", u.Scheme, CIDv1, u.Host),
+				).
+				BodyWithHint(`
+					We return body with HTTP 301 so existing cli scripts that use path-based
+					gateway do not break (curl doesn't auto-redirect without passing -L; wget
```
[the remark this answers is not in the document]
laurentsenta: Yes, I'm just making sure we're preserving the data for now.

tooling/test/config.go
```diff
@@ -0,0 +1,118 @@
...
+	log.Debugf("SubdomainGatewayUrl: %s", SubdomainGatewayUrl)
+	log.Debugf("SubdomainGatewayHost: %s", SubdomainGatewayHost)
+	log.Debugf("SubdomainGatewayScheme: %s", SubdomainGatewayScheme)
+
+	// Generate the localhost subdomain gateway url
+	parsed.Scheme = "http"
+	if parsed.Port() == "" {
+		parsed.Host = "localhost"
+	} else {
+		parsed.Host = fmt.Sprintf("localhost:%s", parsed.Port())
```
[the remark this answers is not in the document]
laurentsenta: Not needed, that's dead code for now, I'll remove it.

tests/t0114_gateway_subdomains_test.go
```diff
@@ -2,61 +2,471 @@ package tests
...
+		))
+
+		with(testGatewayWithManyProtocols(t,
+			"valid file and subdirectory paths in directory listing at {cid}.ipfs.example.com",
+			"{CID}.ipfs.example.com/sub/dir (Directory Listing)",
+			Url("%s://%s.ipfs.%s/", u.Scheme, DirCID, u.Host),
+			Expect().
+				Status(200).
+				Body(And(
+					// TODO: implement html expectations
```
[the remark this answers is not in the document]
laurentsenta: Thanks for the link,
moved to issue: https://github.com/ipfs/gateway-conformance/issues/17

.github/actions/test/action.yml
```diff
@@ -4,6 +4,10 @@ inputs:
   gateway-url:
     description: "The URL of the IPFS Gateway implementation to be tested."
     required: true
+  subdomain-url:
+    description: "The Subdomain URL of the IPFS Gateway implementation to be tested."
+    default: "http://example.com"
+    required: false
```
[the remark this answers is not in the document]
laurentsenta: I'll add a comment in the README, there's no real proxying anymore, we really use that value to overwrite the request sent to the gateway.

.github/workflows/test.yml
```diff
@@ -24,17 +22,25 @@ jobs:
...
         uses: ./gateway-conformance/.github/actions/test
         with:
           gateway-url: http://127.0.0.1:8080
+          subdomain-url: http://example.com
           json: output.json
           xml: output.xml
           html: output.html
           markdown: output.md
-          specs: -subdomain-gateway
+          specs: +subdomain-gateway
```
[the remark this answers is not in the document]
laurentsenta: I think we need to make this feature optional, you need a special configuration on your node to test that feature,
which means messing up your setup even if you're testing something else.

tooling/test/sugar.go
laurentsenta: I'll rename all `Url` names into `URL`, thanks for the catch.

.github/workflows/test.yml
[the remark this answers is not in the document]
laurentsenta: Thinking about it, yours is a better approach; explicit >>> implicit

tooling/test/report.go
```diff
@@ -0,0 +1,94 @@
...
+			
+			var b []byte
+			var err error
+			switch v := v.(type) {
+			case *http.Request:
+				b, err = httputil.DumpRequestOut(v, true)
+			case *http.Response:
+				// TODO: we have to disable the body dump because
+				// it triggers an error:
+				// "http: ContentLength=6 with Body length 0"
```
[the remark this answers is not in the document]
laurentsenta: Not sure yet, I hope to revisit this when we find a case where we really want to see the response in the report.

.github/actions/test/action.yml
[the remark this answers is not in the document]
galargh: I see. Thanks for adding the comment. It makes things clearer. Few more thoughts that it sparked with me.

Wouldn't it make sense to always use what we now call `subdomain-url`. As in, all the test should talk to `gateway-url` in a way that thinks it's called `subdomain-url`? Theoretically, wouldn't that be more correct? So really, every test case should be executed 3 times (with host header, proxy, proxy with tunneling) as long as `subdomain-url` and `gateway-url` are different.

Looking more closely at how we execute the test cases with host header, I think we're not using the protocol part of `gateway-url`. Is that correct?

If so, isn't what we really need from the user these two pieces of info:
- What are you calling/want to call your gateway? Please give us the full URL with a domain name. That would be https://dweb.link for example or http://example.com. You could give us http://127.0.0.1:8080 here as well BUT there's no way for you to support sudomains on that URL so you'll have to disable subdomain-gateway spec. 
- Where is your gateway hosted? Please give us the host we should contact when we're making requests to the URL you provided in the previous step. If you don't, we're just going to use the host from the URL. It's fine if you give us a host with IP or a host with domain. This would be 127.0.0.1 for example or 127.0.0.1:8080 or foo.bar.example.com.

That would come together to something like this - https://github.com/ipfs/gateway-conformance/compare/main...gateway-host (note that I intentionally skipped modifying test/config code; I just wanted to get the idea across). 

There's absolutely no rush with replying to this. I just want to understand what might be most straight-forward way for those running the tests to provide us all the information we need.

tooling/test/report.go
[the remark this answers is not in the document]
galargh: Got it! Thanks :)

PR #11, opened 2023-03-10
chore: try out dorny test-reporter

galargh:


galargh:
It needs to check out the repo with tests - no go for gateway conformance.

PR #15, opened 2023-03-16
feat: turn entrypoint into gateway-conformance CLI

galargh:
I extracted file renames/moves to a separate commit so reviewing commit by commit might make the most sense: https://github.com/ipfs/gateway-conformance/pull/15/commits/c67b7df2dee0806b29c85cb378ca306e617a80e6?diff=split&w=1

Please note that the last commit introduces a different approach to test flags: https://github.com/ipfs/gateway-conformance/pull/15/commits/74b0079ba2375aedaebf351142dbb1f6e9b30de4


tooling/specs/specs.go
```diff
@@ -0,0 +1,72 @@
...
+	default:
+		return false
+	}
+}
+
+type Spec string
+
+const (
+	SubdomainGateway Spec = "subdomain-gateway"
+)
```
galargh, line 29: Maybe these could correspond to the keywords/tag that Robin told us about.

tooling/specs/specs.go
```diff
@@ -0,0 +1,72 @@
...
+
+type maturity string
+
+const (
+	wip        maturity = "wip"
+	draft      maturity = "draft"
+	reliable   maturity = "reliable"
+	stable     maturity = "stable"
+	permanent  maturity = "permanent"
+	deprecated maturity = "deprecated"
```
galargh, line 13: The updated specs specify maturity level per spec: https://github.com/ipfs/specs/pull/382/files#diff-68949a04625be3a1b31a438f35a6e8daaeb881ffd4a1631e083b3b070222e2f9

tooling/specs/specs.go
```diff
@@ -0,0 +1,72 @@
...
+
+var specEnabled = map[Spec]bool{}
+
+func (s Spec) IsEnabled() bool {
+	if enabled, ok := specEnabled[s]; ok {
+		return enabled
+	} else {
+		return s.IsMature()
+	}
+}
```
galargh, line 47: If a spec wasn't enabled/disabled explicitly, we infer whether it's enabled or not based on its' maturity.

tests/init.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (s *specsFlag) Set(value string) error {
+	names := strings.Split(value, ",")
+	var only, enable, disable = []specs.Spec{}, []specs.Spec{}, []specs.Spec{}
+	for _, name := range names {
+		spec, err := specs.FromString(regexp.MustCompile(`^[-+]`).ReplaceAllString(name, ""))
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(name, "+") {
```
galargh, line 25: If a spec is prefixed with `+`, it should be enabled. This can be used to enable tests for specs that are disabled by default (those not mature enough).

For example, let's say spec `fancy` has maturity `wip`. If we run tests without any flags, then tests for `fancy` would be skipped. If we used `-specs=+fancy`, tests for `fancy` would be executed AND all the tests that are enabled by default would be executed too.

tests/init.go
```diff
@@ -0,0 +1,60 @@
...
+			enable = append(enable, spec)
+		} else if strings.HasPrefix(name, "-") {
```
galargh, line 27: If a spec is prefixed with `-`, it should be disabled. This can be used to disable tests for specs that are enabled by default (those mature enough).

For example, we can disable `subdomain-gateway` tests while still testing all the other specs that are enabled by default by providing `-specs=-subdomain-gateway` flag.

tests/init.go
```diff
@@ -0,0 +1,60 @@
...
+			disable = append(disable, spec)
+		} else {
```
galargh, line 29: If a spec is not prefixed with `+` nor `-`, we assume that the user wants to execute only that spec. Using flag `-specs=subdomain-gateway` would disable ALL the specs and then enable `subdomain-gateway` only.

tests/init.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+	}
+	*s = specsFlag(value)
+	return nil
+}
+
+var specsFlagValue specsFlag
+
+func init() {
+	flag.Var(&specsFlagValue, "specs", "comma-separated list of specs to test")
```
galargh, line 59: I implemented this as a single `-specs` flag with differing behaviour based on prefixes but it might as well be three separate flags `-only`, `-enable`, `-disable`. I like the conciseness of `-specs` personally but I definitely can see an argument being made for verbosity of breaking it up.

cmd/gateway-conformance/main.go
```diff
@@ -85,23 +85,18 @@ func main() {
...
-						Name:        "subdomain-gateway-spec",
-						Aliases:     []string{"subdomain-gateway", "subdomain"},
-						Usage:       "Whether the gateway implements the subdomain gateway spec",
-						Value:       true,
-						Destination: &subdomainGatewaySpec,
+					&cli.StringFlag{
+						Name:        "specs",
+						Usage:       "A comma-separated list of specs to test",
+						Value:       "",
+						Destination: &specs,
```
galargh, line 92: I thought about it and I think I'd prefer an interface that people can get used to. If we have separate flag for each spec (as it was here before), then we'd have to change that interface every time a new spec is added to the tests. I think it'd become quite overwhelming with the amount of different specs we might want to support. Also, it'd be harder to maintain as we'd have to remember to update the lists in all these places. In my opinion a single (or three) flags that take a list of specs will be easier to manage in a long run.

galargh: It might be useful to add a command for listing all the specs!

tests/t0114_gateway_subdomains_test.go
```diff
@@ -1,13 +1,11 @@
-//go:build !no_subdomain_gateway_spec
-// +build !no_subdomain_gateway_spec
-
```
galargh: Build tags are cool but I think they might not be powerful enough. E.g. if we had tests that are intersectional, then we'd have to create a new file for them. What I'm saying is, we could probably still achieve similar results but it'd be harder than if we used something that's available during runtime. The additional benefit is that with runtime exclusion, we can choose for the skipped tests to show up as skipped in the report. If we don't build a file, it's as if the tests never existed.

tests/t0114_gateway_subdomains_test.go
```diff
@@ -58,5 +56,7 @@ func TestGatewaySubdomains(t *testing.T) {
 		},
 	}
 
-	Run(t, tests)
+	if SubdomainGateway.IsEnabled() {
+		Run(t, tests)
+	}
```
galargh, line 61: I added it like this for now for simplicity. What I imagine in the future is that each test case comes with a list of specs it tests. Before running the test case, we check if all the specs are enabled. If not, we skip that test case. Additionally, if a test case fails, we could link to the spec in the error message.

To not have to specify specs for each test case, we could wrap the encapsulating array in some class and store the defaults there.

.github/workflows/test.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           xml: output.xml
           html: output.html
           markdown: output.md
-          subdomain-gateway-spec: false
+          specs: -subdomain-gateway
           args: -skip TestGatewayCar
```
galargh, line 38: Disabled it by test name here. I think it's still useful in a case like this. Kubo does support CARs so it wouldn't make sense to disable testing CAR spec altogether. All we want is to disable this single test because Kubo's implementation is wrong at the moment.

tests/init.go
laurentsenta: Should we keep these comments somewhere in the README?

Issue #21, opened 2023-03-24
Add HTTP Expectations


laurentsenta:
See this comment from @darobin, similarly to #17 we should prepare for HTTP expectations as well.

> The one (relatively small) thing I would flag is that you seem to be relying on string matching (equals/contains) but a lot of these values can be wilder (eg. case insensitive, there may be spaces, the value might have to be anchored at the start of the string, etc.). If your HTTP library normalises values, this probably helps a fair bit (worth checking), but it's like that quite a few checks will need to go to regex land. This isn't a big deal and not a problem for the framework, I'm just flagging it as a potential problem that jumped out, probably worth addressing before too many tests are written.

reference: https://github.com/ipfs/gateway-conformance/issues/2#issuecomment-1471971928

## Plan

We should tackle this after all the kubo sharness tests have been ported

- [ ] Look into libraries that might provide similar features (in Go or other languages, we can always reuse their patterns)
- [ ] Then go through the fields, headers, matching, etc. And implement more reliable matchers.

laurentsenta:
Related discussion: https://github.com/ipfs/gateway-conformance/pull/25#discussion_r1154416576

PR #25, opened 2023-03-27
feat: add t0123 & t0113 - gateway - json - cbor - symlinks

laurentsenta:
Contributes to #26 

## Tasks:

- [x] 123: test_dag_pb_conversion
     - [x] Test structure
     - [x] implement dag node getter
     - [x] Implement the body checks
- [x] 123: test_plain_codec
   - [x] Test structure
   - [x] Implement the fixture codec for json
- [x] 123: Pathing, traversal over DAG-JSON and DAG-CBOR
- [x] refactor the fixture into a more usable apis.

## Notes

**bytes equalities**
the current test compares kubo and gateway outputs bytes to bytes, for example: 

```sh
ipfs dag get --output-codec dag-$format $DIR_CID > expected
test_cmp expected my_gateway_output
```

For simple cases, we replace with a higher level comparator (JSON uses `reflect.DeepEqual`),
We stick to a bytes-to-bytes comparison with CBOR for now.

### DIFF
--- a/tooling/car/unixfs.go
+++ b/tooling/car/unixfs.go
@@ -1,5 +1,8 @@
 package car
 
+// Note we force imports of dagcbor, dagjson, and other codecs below.
+// They are registering themselves with the multicodec package
+// during their `init()`.
 import (
 	"bytes"
 	"context"
@@ -23,12 +26,6 @@ import (
 	"github.com/ipld/go-ipld-prime/multicodec"
 	mc "github.com/multiformats/go-multicodec"
 )
-
-func init() {
-	// Note we force imports of dagcbor and dagjson above.
-	// They are registering themselves with the multicodec package
-	// during their `init()`.
-}
 
 type UnixfsDag struct {
 	dsvc  format.DAGService


laurentsenta:
Using shell check, this is what happens on error:

CleanShot 2023-03-30 at 11 01 42@2x

tests/t0122_gateway_tar_test.go
```diff
@@ -69,16 +68,19 @@ func TestGatewayTar(t *testing.T) {
...
+					WithBody("OUTPUT_FILE").
+					With("FILE_CID", fileCID).
+					Succeeds(`
+						cat $OUTPUT_FILE | tar -x;
+						mkdir outputDir;
+						cat $OUTPUT_FILE | tar -x -C outputDir;
+						echo "I am a txt file on path with utf8" > expected
+						test -f outputDir/$FILE_CID;
+						diff -u expected outputDir/$FILE_CID
+						`),
```
galargh, line 82: Could we try to stay Go native? I feel this introduces too big of an opportunity for error. For example, with shell, it's really hard to handle cleanup properly. 

Deferring to shell also ties us to specific system requirements which would be nice to avoid if possible.

It seems to me that with introduction of shell-based checks, we'd get dangerously close to just hiding sharness underneath Go Tests. And, with all the amazing work you've already done with tests, I think that'd be a step backwards.

I'd be happy to try to create a proposal for how we could handle tars "natively". Just to see if that's at all possible, I put together a Go tar reader which seems to enable exactly this - https://gist.github.com/galargh/7a1baec0f679605e0ba3ecc0d7e558fc

laurentsenta: Agreed, I don't want the shell to become a new default in the tests either, but I want to be pragmatic about it: we can use some to help us bridge the gap between the sharness and this test suite.

After all, we are migrating a shell-based test suite to Go, so we're not adding "dirty" shell code, but rather removing it.

Each PR has to accomplish a lot:

- The tedious work of porting the tests from sharness to the test suite.
- Some refactoring of the tests to make them nicer and less awkward (like merging cases when we test multiple headers on one request),
- And then implementing new APIs.

Implementing a new API, especially for tar, is not very hard, but it does introduce scope creep in each PR, and we have to anticipate the use cases in advance, even for a single small test.

I would like to use shell to avoid solving three problems with each PR. Once we have a complete test suite, we can go back and make native what needs to be native.

As for the specific issue of cleanup, the checker does take care of it.

galargh: I don't agree that tar handling is a case that warrants adding a whole new, extremely powerful API which interacts with shell directly.

If we don't want to work on adding new APIs at the same time as porting tests, we could skip adding tests that cannot be handled with the current APIs and document them instead. That would give us an overview of what APIs are missing at the end of the process.

To reduce the number of things we're doing in the test port PRs, I think we could also skip the consolidation for now.

galargh: > I'm worried we spend time implementing one-time use APIs instead of working on critical stuff like IPNS and implementing spec-related tests & features.

If it's taking us away from more critical work, then I think we shouldn't be spending time on trying to handle those test cases right now at all.

galargh: Tackling fewer input tests per PR might also help reduce scope.

galargh: One more thought - the tar handling doesn't really have to become API straight away. It could be a custom inline check  too. Something like:
```
Response().Body().Check(func (body byte[]) error {
  // check if the body is the tar we expect
})
```

That would remove the need for trying to guess the interface in advance while not having to resort to interacting with shell.

laurentsenta: I created a PR in https://github.com/ipfs/gateway-conformance/pull/28 to review 122 specifically.

> If we don't want to work on adding new APIs at the same time as porting tests, we could skip adding tests that cannot be handled with the current APIs and document them instead. That would give us an overview of what APIs are missing at the end of the process.

It really depends on our goal, between tradeoffs I'd pick the one where we get more tests from more contributors, so:

between:
- spending time implementing and maintaining 50 lines of Go code that will only be used once, and redoing it for every missing API,
- or leaving todo comments in the code,

We benefit more from having a test that runs thanks to a few lines of shell script. It's an intermediate solution that we can reuse when two lines of bash do the job.

And that's true for external contributors as well: right now, if you want to test a behavior not covered by the API, even just to create a repro and propose a PR, you have to be pretty familiar with golang and the PLABs stack.

fixtures/t0123/dag-pb.json
galargh, line 1: Are we using this file anywhere? I cannot find it.

tests/t0123_gateway_json_cbor_test.go
```diff
@@ -1,13 +1,858 @@
...
+			test_cmp cat_output curl_output &&
+			test_should_contain "Content-Type: text/plain" headers &&
+			test_should_not_contain "Content-Type: application/$format" headers &&
+			test_should_not_contain "Content-Type: application/vnd.ipld.dag-$format" headers
+			'
+			*/
+			{
+				Name: fmt.Sprintf("GET UnixFS with format=%s (not dag-%s) is no-op (no conversion)", row.Format, row.Format),
+				Request: Request().
+					Path("ipfs/%s?format=%s", fileCID, row.Format),
```
galargh, line 208: Is there are reason to use one version over another or are they interchangeable?
```suggestion
					Path("ipfs/%s", fileCID).
					Query("format", row.Format),
```

tests/t0123_gateway_json_cbor_test.go
```diff
@@ -1,13 +1,858 @@
...
+				Hint: `(basic checks, detailed behavior is tested in t0116-gateway-cache.sh)`,
+				Request: Request().
+					Path("ipfs/%s", dagTraversalCID).
+					Header("Accept", fmt.Sprintf("application/vnd.ipld.dag-%s", row.Format)),
+				Response: Expect().
+					Headers(
+						Header("Etag").Hint("includes Etag").Contains("%s.dag-%s", dagTraversalCID, row.Format),
+						Header("X-Ipfs-Path").Hint("includes X-Ipfs-Path").Exists(),
+						Header("X-Ipfs-Roots").Hint("includes X-Ipfs-Roots").Exists(),
+						Header("Cache-Control").Hint("includes Cache-Control").Contains("public, max-age=29030400, immutable"),
```
galargh, line 725: TODO for later, we're going to want to go through all those exact cache-control checks and make them more lenient most likely.

tooling/car/unixfs.go
```diff
@@ -1,21 +1,35 @@
...
+	_ "github.com/ipld/go-ipld-prime/codec/json"
+	"github.com/ipld/go-ipld-prime/multicodec"
+	mc "github.com/multiformats/go-multicodec"
 )
 
+func init() {
+	// Note we force imports of dagcbor and dagjson above.
+	// They are registering themselves with the multicodec package
+	// during their `init()`.
+}
```
galargh, line 31: Do we still need the init here?

tooling/check/check.go
```diff
@@ -196,3 +247,72 @@ func (c CheckFunc[T]) Check(v T) CheckOutput {
 }
 
 var _ Check[string] = &CheckFunc[string]{}
+
+type CheckNot struct {
+	check Check[string]
+}
+
+func Not(check Check[string]) Check[string] {
```
galargh, line 255: Nice 🍭

galargh (approved):
🏆

tooling/car/unixfs.go
laurentsenta: Fixed, moved to a comment above the imports

tests/t0123_gateway_json_cbor_test.go
```diff
@@ -1,13 +1,859 @@
...
+					),
+			},
+			/**
+			  test_expect_success "GET $name from /ipfs for application/$format returns the same payload as format=dag-$format" '
+			  curl -sX GET "http://127.0.0.1:$GWAY_PORT/ipfs/$CID?format=dag-$format" -o expected &&
+			  curl -sX GET "http://127.0.0.1:$GWAY_PORT/ipfs/$CID?format=$format" -o plain_output &&
+			  test_cmp expected plain_output
+			  '
+			*/
+			// TODO(lidel): Note we disable this test, we check the payloads above.
```
laurentsenta, line 567: @lidel a note here, we ignore a few tests: they compare two requests with each other.
Instead, we test each request.

tests/t0123_gateway_json_cbor_test.go
```diff
@@ -1,13 +1,859 @@
...
+			Name: "GET DAG-CBOR traverses multiple links",
+			Request: Request().
+				Path("ipfs/%s/foo/link/bar", dagCBORTraversalCID).
+				Query("format", "dag-json"),
+			Response: Expect().
+				Status(200).
+				Body(
+					// TODO: I like that this text is readable and easy to understand.
+					// 		 but we might prefer matching abstract values, something like "IsJSONEqual(someFixture.formatedAsJSON))"
+					IsJSONEqual([]byte(`{"hello": "this is not a link"}`)),
```
laurentsenta, line 490: @galargh I'll keep as is, sounds good?

galargh: Would be nice if we could retrieve it from the car, but sounds good to me with the TODO 👍

fixtures/t0123/dag-pb.json
laurentsenta: Good catch! I removed it.

tests/t0123_gateway_json_cbor_test.go
laurentsenta: 👍 we have https://github.com/ipfs/gateway-conformance/issues/21 to capture this.

PR #28, opened 2023-03-31
feat: add t0122 - tar

laurentsenta:


laurentsenta:
(moving into its own PR to simplify discussion about https://github.com/ipfs/gateway-conformance/pull/25#discussion_r1153006764)

galargh (changes requested):
Following up on https://github.com/ipfs/gateway-conformance/pull/25#discussion_r1154293484

With the addition of shell, we'd introduce a whole world of undefined dependencies, we'd increase the bar for understanding tests because now you also have to be proficient with bash, and we'd add a whole lot of code that's very hard to test. I'd consider all of these to be discouraging for contributors.

One of the core goals of the project is to make the tests more reliable. In my opinion, this would defeat this purpose completely.

galargh (approved):

PR #29, opened 2023-03-31
feat: add a randomizer test

laurentsenta:
The worst thing that could happen to us is introducing a bug that disables one or more checks.

An example: test t118 in Kubo, due to a typo, we miss a bug that exists in Kubo
https://github.com/ipfs/kubo/pull/9652/files

It's not crazy to imagine something similar could happen here, for one test or more.
The suite would be green, but the code wouldn't be tested.

We also need extra care when we port kubo tests: we can't do red/green testing.
So a test writer I have to introduce bugs (add typos to asserts or negate values) to make sure the asserts are correct.

This PR proposes a "randomizer", which lives in front of a gateway and randomizes its outputs (headers and body).

If we run the test suite against our randomizer, we expect every test to fail most of the time.
If the entire test suite passes, there is a problem.

- This is the first step towards "testing the test suite"; it requires little code, but we can validate that we catch the most significant errors locally and in the CI.
- We implicitly get the benefits of red/green testing, because we now that tests are actually testing things.


laurentsenta:
@lidel @galargh I'd like to get your feedback on that one,
it doesn't replace unit-tests but having a no-maintenance-catch-big-mistakes test would help me sleep better.

galargh:
I like it! That's really clever ❤️

galargh:
Maybe we could run tests against random gateway a couple of times. Or rerun the tests that passed. Then, if we found a test case that passed on all the attempts, we'd know something's wrong.

PR #30, opened 2023-04-03
wip: feat/t0116 gateway cache.sh

galargh:
This PR ends up doing a bit more than I intended at first. It:
- rewrites t0116 in sugar mode - https://github.com/ipfs/gateway-conformance/pull/30/commits/1cf17297ef19ff7fac17ffc935fa5c428ac80d71
- makes `SugarTests` the default for defining tests - https://github.com/ipfs/gateway-conformance/pull/30/commits/5dde1751f204c2c4c2f2847851f3c3bc8806c9c1
- makes `SugarTests` the only option for defining tests - https://github.com/ipfs/gateway-conformance/pull/30/commits/72453bcd1a761f40530c2e6c831d08c25a321f63 and https://github.com/ipfs/gateway-conformance/pull/30/commits/cdddcc91fd555002a125613f618d9f927d825b22
- adds `Provider` interface (think `StringProvider(value).Get() rather than using value directly) - https://github.com/ipfs/gateway-conformance/pull/30/commits/284e1e62efa5f4253586dade6790f3b5a645ccb3
- adds a way to init header value from a variable - https://github.com/ipfs/gateway-conformance/pull/30/commits/8e967b3ae094f03b3ade0dad5271195f396d1493
- adds a way to init header value with a function - https://github.com/ipfs/gateway-conformance/pull/30/commits/f5cace99713711d0ffb5114c7e6b07066951f014
- enables t0116 test cases that depend on responses from other test cases - https://github.com/ipfs/gateway-conformance/pull/30/commits/4e7e0486addcb1ed1c59b034d4f0fdae7816eebe
- adds comments on next steps - https://github.com/ipfs/gateway-conformance/pull/30/commits/2f9adafacfedf8749a7f6e8e0b5675854f165e50

I mentioned some of the interesting commits up here. I think looking at it that way is much more convenient. We can definitely discuss this in person too!

Feedback is more than welcome. 

I think embracing SugarTests would let us simplify quite a bit of the internals. 

The way I implemented information sharing between requests... I'm not 100% convinced. 
- It works cause' test cases are executed sequentially, which is always true for sub-test cases. 
- If we were to exclude the dependency test case or run only the test that has dependencies, it'd fail. It might be fine because those are power user options anyway. 
- If the dependency test case failed, the dependee would fail too, but the error wouldn't indicate the reason too well. Maybe it's enough to check the request before making it and add some hints around it.

Other things for consideration:
- We could group tests with dependencies more formally so that if the dependency fails, we don't even execute the latter test cases.
- The request from the dependency could be done outside test cases. The downside is, any failure would hold back all the test cases.

Finally, important to note that this PR doesn't enable IPNS test cases.

### DIFF
--- a/tests/t0113_gateway_symlink_test.go
+++ b/tests/t0113_gateway_symlink_test.go
@@ -11,24 +11,22 @@ import (
 func TestGatewaySymlink(t *testing.T) {
 	fixture := car.MustOpenUnixfsCar("t0113-gateway-symlink.car")
 
-	tests := []CTest{
+	tests := SugarTests{
 		{
 			Name: "Test the directory listing",
 			Request: Request().
-				Path("ipfs/%s?format=raw", fixture.MustGetCid()).Request(),
+				Path("ipfs/%s?format=raw", fixture.MustGetCid()),
 			Response: Expect().
 				Status(200).
-				Body(fixture.MustGetRawData()).
-				Response(),
+				Body(fixture.MustGetRawData()),
 		},
 		{
 			Name: "Test the symlink",
 			Request: Request().
-				Path("ipfs/%s/bar", fixture.MustGetCid()).Request(),
+				Path("ipfs/%s/bar", fixture.MustGetCid()),
 			Response: Expect().
 				Status(200).
-				Bytes("foo").
-				Response(),
+				Bytes("foo"),
 		},
 	}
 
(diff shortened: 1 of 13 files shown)


laurentsenta:
(removing myself while this is still in draft)

galargh:
For what it's worth, it's never getting out of draft but I'm keeping it alive to remember to salvage parts of it that are still relevant.

Issue #26, opened 2023-03-27
Port all "simple" sharness tests


laurentsenta:
#3 relies on #22 and #10 which requires new features / larger tasks.

In parallel, we'll push more tests to the test suite to support bifrost

- [x] t0109-gateway-web-_redirects.sh (@lidel's note: afaik fixtures come from https://github.com/ipfs/specs/pull/290)
    - [x] basics: https://github.com/ipfs/gateway-conformance/pull/33
    - [x] dnslink: https://github.com/ipfs/gateway-conformance/pull/34
- [x] ~t0111-gateway-writeable.sh~  - no longer a thing (https://github.com/ipfs/kubo/pull/9743)
- [x] t0112-gateway-cors.sh - https://github.com/ipfs/gateway-conformance/pull/52
    - @lidel's note: Kubo returns multiple headers instead of array, tests should support both
- [x] t0113-gateway-symlink.sh
    - (@lidel's note: this ive very uncharted teritory, if anything weird is discovered here, comment on https://github.com/ipfs/specs/issues/368 :pray: )
    - #25 
- [x] t0114-gateway-subdomains.sh
    - [x] basics: https://github.com/ipfs/gateway-conformance/pull/9
    - [x] dnslink: https://github.com/ipfs/gateway-conformance/pull/35
    - [x] ipns + dnslink: https://github.com/ipfs/gateway-conformance/pull/53
- [x] t0115-gateway-dir-listing.sh - https://github.com/ipfs/gateway-conformance/pull/58
- [x] t0122-gateway-tar.sh #28 
- [x] t0123-gateway-json-cbor.sh
    - [x] non-ipns part - #25
    - [x] ipns part - https://github.com/ipfs/gateway-conformance/pull/59

Other tests implemented along other tickets:

- [x] ~t0117-gateway-block.sh~ #3
- [x] ~t0118-gateway-car.sh~ #3
- [x] ~t0124-gateway-ipns-record.sh~ #3
- [x] ~t0400-api-no-gateway.sh~ - kubo specific

## follow-up tasks

- [x] t0116-gateway-cache.sh  (@lidel's note: pretty important one) https://github.com/ipfs/gateway-conformance/pull/97

galargh:
@laurentsenta I'm going to try to pick up `t0116-gateway-cache.sh`. I'll report at the end of the day what the status is.

BigLep:
[...]

PR #33, opened 2023-04-05
feat: add test 109

laurentsenta:
Contributes to https://github.com/ipfs/gateway-conformance/issues/26


tests/t0109_gateway_web_redirects_test.go
```diff
@@ -0,0 +1,391 @@
...
+							Contains("could not parse _redirects:"),
+							Contains("redirects file size cannot exceed"),
+						),
+					),
+			},
+		}...)
+	}
+
+	if specs.SubdomainGateway.IsEnabled() {
+		Run(t, unwrapTests(t, tests).Build())
```
laurentsenta, line 308: @galargh I came up with a new pattern for subdomain:
instead of explicitly calling a function on every test:
https://github.com/ipfs/gateway-conformance/blob/023888a768b3d9d639fec2ed5e898d52de92b573/tests/t0114_gateway_subdomains_test.go#L49-L56

`unwrapTests` takes a bunch of tests relying on a subdomain, and unwrap them, it's much more natural when you write the tests.

Not perfect, but a step in a better direction I think, maybe that'll some ideas on your side :)

tooling/car/merge.go
```diff
@@ -4,8 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/ipfs/boxo/ipld/car/v2/blockstore"
 	"github.com/ipfs/go-cid"
-	"github.com/ipld/go-car/v2/blockstore"
```
laurentsenta, line 8: @galargh that's the fix for the crazy bug I had before,
- I double-checked: `init()` is called only once per package as we expected. 
- When I renamed `fixture.go` into `zfixture.go`, the bug was fixed 🎉
- I realized this had to be a collision between new boxo packages and the old ones.

galargh: 🤯

PR #34, opened 2023-04-05
feat: add dnslink support and update t0109

### LINKED ISSUE #22: Add dnslink support
To complete porting the 114 test, we also need to support dns-link fixture.

laurentsenta:
Contributes to #26 
Closes #22

This tests adds dnslink support:

We introduce a fixture type which contains a dnslink configuration:

```yml
dnslinks:
  custom-dnslink:
    subdomain: dnslink-enabled-on-fqdn
    path: /ipfs/QmYBhLYDwVFvxos9h8CGU2ibaY66QNgv8hpfewxaQrPiZj
```

Test writer uses the fixture with:

```go
dnsLinks := dnslink.MustOpenDNSLink("t0109-dnslink.yml")
dnsLink := dnsLinks.Get("custom-dnslink")

// then use it in your subdomain gateway URL test:
gatewayURL := SubdomainGatewayURL
u, _ := url.Parse(gatewayURL)

// dnsLinkBaseUrl = "http://dnslink-enabled-on-fqdn.foo.bar.example.com"
dnsLinkBaseUrl := fmt.Sprintf("%s://%s%s", u.Scheme, dnsLink, u.Host)
```

Test user calls the `export` CLI to get the yml files. We also generate a dnslinks.json.

There is a demo in `kubo-config.example.sh` that shows how to import the fixtures:

```bash
IPFS_NS_MAP=$(cat ./dnslinks.json | jq -r 'to_entries | map("\(.key).example.com:\(.value)") | join(",")')

echo "Set the following IPFS_NS_MAP before starting the kubo daemon:"
echo "IPFS_NS_MAP=${IPFS_NS_MAP}"
```

## Todo

- [x] Clarify the API and how we want to expose this to users
- [x] helper libs reusable
- [x] finalize in CI (import the env when the daemon starts)


laurentsenta:
failing test on purpose: https://github.com/ipfs/gateway-conformance/actions/runs/4924297430

fixtures/t0109-dnslink.yml
```diff
@@ -0,0 +1,6 @@
+dnslinks:
```
galargh, line 1: Should we add a JSON schema for the config files?

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "dnslinks": {
      "type": "object",
      "additionalProperties": {
        "type": "object",
        "properties": {
          "subdomain": {
            "type": "string"
          },
          "path": {
            "type": "string"
          }
        },
        "required": ["subdomain", "path"],
        "additionalProperties": false
      }
    }
  },
  "additionalProperties": false
}
```

Then, there are editor plugins that could verify the config files for us if we provide the path to the schema like this:

```suggestion
# yaml-language-server: $schema=PATH_TO_SCHEMA_DOT_JSON
dnslinks:
```

kubo-config.example.sh
```diff
@@ -9,4 +9,9 @@ ipfs config --json Gateway.PublicGateways '{
 		"InlineDNSLink": true,					
 		"Paths": ["/ipfs", "/ipns", "/api"]		
 	}											
-}'
\ No newline at end of file
+}'
+
+IPFS_NS_MAP=$(cat ./dnslinks.json | jq -r 'to_entries | map("\(.key).example.com:\(.value)") | join(",")')
```
galargh, line 14: If we export, then I think we could call this script as `source kubo-config.example.sh` to get the env var in parent shell.

```suggestion
export IPFS_NS_MAP=$(cat ./dnslinks.json | jq -r 'to_entries | map("\(.key).example.com:\(.value)") | join(",")')
```

galargh (approved):
Nothing blocking 🥇

fixtures/t0109-dnslink.yml
laurentsenta: I love it, added

tooling/dnslink/dnslink.go
```diff
@@ -0,0 +1,52 @@
...
+	fixturePath := path.Join(fixtures.Dir(), file)
+	dnsLinks, err := OpenDNSLink(fixturePath)
+	if err != nil {
+		panic(err)
+	}
+
+	return dnsLinks
+}
+
+func (d *DNSLinks) Get(id string) string {
```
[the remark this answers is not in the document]
laurentsenta: Good catch, thanks

kubo-config.example.sh
laurentsenta: 👍 good catch, that should make the CI test simpler, I'll update the script.

fixtures/t0109-dnslink.yml
```diff
@@ -0,0 +1,7 @@
+# yaml-language-server: $schema=fixture.schema.json
+dnslinks:
+  custom-dnslink:
+    subdomain: dnslink-enabled-on-fqdn
```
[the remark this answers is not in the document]
laurentsenta: > (github was down and missed oportunity to review earlier)

👍  thanks for reviewing anyway; happy to fix if the approach is lacking, @hacdias had a similar question when we reviewed it.

> DNSLink TXT records can be present on subdomains OR the top domains right after TLD.

My understanding is that we want to configure the top domain + tld during test EXECUTION, not at test definition: two users might run that test on different subdomain gateway URLs, `example.com` and `dweb.link` for example.

This means we want to be able to construct URLs with `{dns-link.subdomain}.{gateway_url}` at run time ([actual code](https://github.com/ipfs/gateway-conformance/blob/9e419ec14a86c0c5c8b6b41b40170b95026f63f3/tests/t0109_gateway_web_redirects_test.go#L320-L326)).

Thinking about it, I guess we might want to define a fqdn/domain, like
```yaml
dnslinks:
    wikipedia:
        domain: wikipedia-on-ipfs.org
        path: /ipfs/somethingsomething
```

but that would be a different test, we do need to keep the `subdomain` option to port most of the tests like [114](https://github.com/ipfs/kubo/blob/0e7331c952b40d481efc717b0d749af2bd6bb92f/test/sharness/t0114-gateway-subdomains.sh#L557-L566) and [109](https://github.com/ipfs/kubo/blob/0e7331c952b40d481efc717b0d749af2bd6bb92f/test/sharness/t0109-gateway-web-_redirects.sh#L182-L200).

Issue #10, opened 2023-03-08
Add ipns record support


laurentsenta:
## ipns record in tests:

```golang
ipnsRecord := ipns.MustOpenRecord("t0124/key1.ipns-record")

ipnsRecord.TTL() // used to validate the cache output from API
ipnsRecord.Verify(bytes) // used to validate the output from API
ipnsRecord.Name() // used to contruct /ipfs/NAME
```

Also add means to check the `Body` output:
https://github.com/ipfs/kubo/blob/master/test/sharness/t0124-gateway-ipns-record.sh

## ipns record during provision

- [ ] export the records
- [ ] use them during CI testing

laurentsenta:
https://specs.ipfs.tech/ipns/ipns-record/#ipns-record

the public key might not live in the record:
> If public key is small enough to fit in IPNS name (e.g., Ed25519 keys inlined using identity multihash), IpnsEntry.pubKey field is redundant and may be skipped to save space.

Which means we need to move around the public key / ipns name along with the record. File name is fine, but how do you store a record for an invalid key?
- [key]_suffix.ipns-record
- or just move in a `.yml` file like we do currently with dnslink?

laurentsenta:
- [ ] At the moment, when I open an ipns record, the `PublicKey` is equal to `"m"`, why?

laurentsenta:
- [ ] follow-up task: once we have ported the sharness tests, take some time to discuss what we need from ipns records
    - [ ] embed them in a car file served by a cargateway? 
    - [ ] do we want to keep invalid ipns records?
    - [ ] how to store & use them in a way that is easier to maintain (right now the name has to contain the peerId)

laurentsenta:
Complete, see the PR for discussion about car serving, etc.

PR #37, opened 2023-04-10
feat: add t0124 with ipns fixtures

laurentsenta:
contributes to https://github.com/ipfs/gateway-conformance/issues/10
requires https://github.com/ipfs/kubo/pull/9667 (allow-offline for ipns fixtures)

🎶 the regular test is failing; this is expected: we need the `ipfs routing put --allow-offline` option that is not released yet. The second workflow "Test Master" is passing.

[Failing test](https://github.com/ipfs/gateway-conformance/actions/runs/4948360492/jobs/8848968179) to make sure tests are running - we removed the ipns-record on purpose.

- [x] implement ipns-record fixture loader & exporter
- [x] implement ipns checker for response
- [x] t0116-gateway-cache.sh
- [x] t0124-gateway-ipns-record.sh
- [x] fix fixture collision names
- [x] add a test runner on kubo's master

## Todo

- [x] kubo: make it possible to import records even in offline mode
- [x] kubo: make it possible to call the api in offline mode
    - If we try to get the record with `http://127.0.0.1:8080/ipns/k51qzi5uqu5dhjjqhpcuvdnskq4mz84a2xg1rpqzi6s5460q2612egkfjsk42x?format=ipns-record`
the call timeouts because kubo take its time to load the record from the network.
    - If we try to start the daemon in  offline mode we get "this action must be run in online mode, try running 'ipfs daemon' first".
- [x] Discuss code reuse with the team, we have to copy and paste a lot of code (see kubo_name)
- [x] validate the API:
    - [x] how we distribute ipns-records
    - [x] how we export as fixture & use them in tests 

## Follow-up

- [ ] t0114-gateway-subdomains.sh
- [ ] t0123-gateway-json-cbor.sh


laurentsenta:
(temporary target on t0109 because this builts on top of dnslink)

laurentsenta:
(note the test will not pass until https://github.com/ipfs/kubo/pull/9667 is released)

cmd/gateway-conformance/main.go
```diff
@@ -190,6 +190,12 @@ func main() {
 						if err != nil {
 							return err
 						}
+
+						// TODO: merge IPNS records somehow
+						err = copyFiles(fxs.IPNSRecords, directory)
```
lidel, line 195: :thought_balloon: Some related thoughts/ideas in https://github.com/ipfs/specs/issues/369.
(tldr fixtures could be included as blocks in a CAR with immutable data, as blocks with `ipns-record` codec)

tooling/ipns/kubo_name.go
```diff
@@ -0,0 +1,147 @@
+/**
+ * This was copied from Kubo's `name` command
+ * TODO: discuss with the team about setting up a reusable abstraction somewhere.
```
lidel, line 3: :thought_balloon: we've moved go-ipns to [`boxo/ipns`](https://github.com/ipfs/boxo/tree/main/ipns), my vote would be to move these utility functions there, and avoid duplicated code.

This cleanup could be a part of https://github.com/ipfs/specs/issues/376,  or something you  do before.

BigLep:
@laurentsenta : please request review from @hacdias when this is out of draft.

BigLep:
@laurentsenta : I saw the comment [here](https://www.notion.so/pl-strflt/Laurent-5a1f9c6b74d143d49c7b5adc943a3951?pvs=4#d11aa2dcd41d45aaa3b7dd88b8b8caa0) about waiting on review.  Who are you waiting on review from?  This may not be clear given that this is in draft.

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,115 @@
...
+	file := fixture.MustGetRoot()
+	fileCID := file.Cid()
+
+	// test_expect_success "Add the test directory & IPNS records" '
+	// ipfs dag import ../t0124-gateway-ipns-record/fixtures.car &&
+	// ipfs routing put /ipns/${IPNS_KEY} ../t0124-gateway-ipns-record/${IPNS_KEY}.ipns-record
+	// '
+	// TODO: what if you'd import with `t0124/simple-id.ipns-record`,
+	// then the loader "just" finds the file with `peer_id_simple-id.ipns-record`?
+	ipns := ipns.MustOpenIPNSRecordWithKey("t0124/k51qzi5uqu5dh71qgwangrt6r0nd4094i88nsady6qgd1dhjcyfsaqmpp143ab.ipns-record")
```
hacdias, line 24: Is this `TODO` still relevant? Or is "simple ID" something else?

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,115 @@
...
+	ipnsKey := ipns.Key()
+
+	tests := SugarTests{
+		// test_expect_success "Create and Publish IPNS Key" '
+		// curl "http://127.0.0.1:$GWAY_PORT/ipns/$IPNS_KEY" > curl_output_filename &&
+		// test_should_contain "Hello IPFS" curl_output_filename
+		// '
+		{
+			Name: "Create and Publish IPNS Key",
```
hacdias, line 33: I would rename this. We're neither creating, nor publishing an IPNS Key. Perhaps `Gateway Resolves IPNS Key` or something along those lines. We're just checking if the key is resolved to the right content.

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,115 @@
...
+		// test_should_contain "$FILE_CID" verify_output
+		// '
+		// test_expect_success "GET KEY with format=ipns-record has expected HTTP headers" '
+		// curl -sD - "http://127.0.0.1:$GWAY_PORT/ipns/$IPNS_KEY?format=ipns-record" > curl_output_filename 2>&1 &&
+		// test_should_contain "Content-Disposition: attachment;" curl_output_filename &&
+		// test_should_contain "Content-Type: application/vnd.ipfs.ipns-record" curl_output_filename &&
+		// test_should_contain "Cache-Control: public, max-age=3155760000" curl_output_filename
+		// '
+		{
+			Name: "GET KEY with format=ipns-record and validate key",
```
hacdias, line 51:
```suggestion
			Name: "GET KEY with format=ipns-record has expected HTTP headers and valid key",
```

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,115 @@
...
+		// test_should_contain "$FILE_CID" verify_output
+		// '
+		// test_expect_success "GET KEY with 'Accept: application/vnd.ipfs.ipns-record' has expected HTTP headers" '
+		// curl -H "Accept: application/vnd.ipfs.ipns-record" -sD - "http://127.0.0.1:$GWAY_PORT/ipns/$IPNS_KEY" > curl_output_filename 2>&1 &&
+		// test_should_contain "Content-Disposition: attachment;" curl_output_filename &&
+		// test_should_contain "Content-Type: application/vnd.ipfs.ipns-record" curl_output_filename &&
+		// test_should_contain "Cache-Control: public, max-age=3155760000" curl_output_filename
+		// '
+		{
+			Name: "GET KEY with 'Accept: application/vnd.ipfs.ipns-record' and validate key",
```
hacdias, line 79:
```suggestion
			Name: "GET KEY with f'Accept: application/vnd.ipfs.ipns-record' has expected HTTP headers and valid key",
```

tooling/ipns/ipns.go
```diff
@@ -0,0 +1,87 @@
...
+// 	return dnsLinks
+// }
+
+// func (d *DNSLinks) Get(id string) string {
+// 	dnsLink, ok := d.DNSLinks[id]
+// 	if !ok {
+// 		panic(fmt.Errorf("dnslink %s not found", id))
+// 	}
+// 	return dnsLink.Subdomain
+// }
```
hacdias, line 87: If not used, I would suggest removing before landing this. This also goes for remaining commented code that is not a missing test.

hacdias (commented):
Very nice overall. I also wanted to ask if there is any reason why we're keeping the bash code from the old sharness tests as comments. I would personally like to get them cleanup up, but I just want to understand the reasoning. Is it to make it clear which tests are which during the transitioning period from sharness to the gateway conformance suite?

tests/t0124_gateway_ipns_record_test.go
laurentsenta: Thank for the check, it's outdated, I learned later that we have to keep the pubkey around since it's not always in the ipns-record itself.

laurentsenta:
Thanks for the review @hacdias & for the note @BigLep,

We had a sync this morning with @galargh and @hacdias we reviewed and agreed on the API, with #34 merged I will move this one out of draft next.

hacdias:
@laurentsenta this is how I would update to use the new boxo/ipns.ValidateWithPeerID. Sadly, I can't commit here, so here it is:

```diff
diff --git a/go.mod b/go.mod
index 1d7b8f0..47500e3 100644
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/ipfs/gateway-conformance
 go 1.20
 
 require (
-	github.com/ipfs/boxo v0.8.1
+	github.com/ipfs/boxo v0.8.2-0.20230510114019-33e3f0cd052b
 	github.com/ipfs/go-cid v0.4.1
 	github.com/ipld/go-ipld-prime v0.20.0
 	github.com/libp2p/go-libp2p v0.26.3
diff --git a/go.sum b/go.sum
index bfa7c29..cf7a061 100644
--- a/go.sum
+++ b/go.sum
@@ -47,8 +47,8 @@ github.com/hashicorp/golang-lru v0.5.4/go.mod h1:iADmTwqILo4mZ8BN3D2Q6+9jd8WM5uG
 github.com/huin/goupnp v1.0.3 h1:N8No57ls+MnjlB+JPiCVSOyy/ot7MJTqlo7rn+NYSqQ=
 github.com/ipfs/bbloom v0.0.4 h1:Gi+8EGJ2y5qiD5FbsbpX/TMNcJw8gSqr7eyjHa4Fhvs=
 github.com/ipfs/bbloom v0.0.4/go.mod h1:cS9YprKXpoZ9lT0n/Mw/a6/aFV6DTjTLYHeA+gyqMG0=
-github.com/ipfs/boxo v0.8.1 h1:3DkKBCK+3rdEB5t77WDShUXXhktYwH99mkAsgajsKrU=
-github.com/ipfs/boxo v0.8.1/go.mod h1:xJ2hVb4La5WyD7GvKYE0lq2g1rmQZoCD2K4WNrV6aZI=
+github.com/ipfs/boxo v0.8.2-0.20230510114019-33e3f0cd052b h1:6EVpfwbBgwhfZOA19i55jOGokKOy+OaQAm1dg4RbXmc=
+github.com/ipfs/boxo v0.8.2-0.20230510114019-33e3f0cd052b/go.mod h1:Ej2r08Z4VIaFKqY08UXMNhwcLf6VekHhK8c+KqA1B9Y=
 github.com/ipfs/go-bitfield v1.1.0 h1:fh7FIo8bSwaJEh6DdTWbCeZ1eqOaOkKFI74SCnsWbGA=
 github.com/ipfs/go-bitfield v1.1.0/go.mod h1:paqf1wjq/D2BBmzfTVFlJQ9IlFOZpg422HL0HqsGWHU=
 github.com/ipfs/go-block-format v0.0.2/go.mod h1:AWR46JfpcObNfg3ok2JHDUfdiHRgWhJgCQF+KIgOPJY=
diff --git a/tooling/ipns/ipns_test.go b/tooling/ipns/ipns_test.go
index 4cced6f..b2c9dbb 100644
--- a/tooling/ipns/ipns_test.go
+++ b/tooling/ipns/ipns_test.go
@@ -42,9 +42,8 @@ func TestLoadIPNSRecord(t *testing.T) {
 	assert.Equal(t, ipns.Value(), "/ipfs/bafkreidfdrlkeq4m4xnxuyx6iae76fdm4wgl5d4xzsb77ixhyqwumhz244")
 	assert.True(t, ipns.Entry.Validity.After(YEAR_100))
 
-	verify, err := ipns.Verify()
-	assert.Nil(t, err)
-	assert.True(t, verify)
+	err = ipns.Valid()
+	assert.NoError(t, err)
 }
 
 func TestLoadTestRecord(t *testing.T) {
@@ -60,7 +59,6 @@ func TestLoadTestRecord(t *testing.T) {
 	assert.Equal(t, ipns.Value(), "/ipfs/bafyaaeykceeaeeqlnbswy3dpo5xxe3debimaw")
 	assert.True(t, ipns.Entry.Validity.After(YEAR_100))
 
-	verify, err := ipns.Verify()
-	assert.Nil(t, err)
-	assert.True(t, verify)
+	err = ipns.Valid()
+	assert.NoError(t, err)
 }
diff --git a/tooling/ipns/kubo_name.go b/tooling/ipns/kubo_name.go
index b97fc9c..0cf288e 100644
--- a/tooling/ipns/kubo_name.go
+++ b/tooling/ipns/kubo_name.go
@@ -7,7 +7,6 @@ package ipns
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"time"
 
 	"github.com/gogo/protobuf/proto"
@@ -16,7 +15,6 @@ import (
 	"github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/codec/dagcbor"
 	"github.com/ipld/go-ipld-prime/codec/dagjson"
-	ic "github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/peer"
 	mbase "github.com/multiformats/go-multibase"
 )
@@ -24,15 +22,15 @@ import (
 // IpnsInspectEntry contains the deserialized values from an IPNS Entry:
 // https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-serialization-format
 type IpnsInspectEntry struct {
-	Value        string                     `json:"value"`
+	Value        string                          `json:"value"`
 	ValidityType *ipns_pb.IpnsEntry_ValidityType `json:"validityType"`
-	Validity     *time.Time                 `json:"validity"`
-	Sequence     uint64                     `json:"sequence"`
-	TTL          *uint64                    `json:"ttl"`
-	PublicKey    string                     `json:"publicKey"`
-	SignatureV1  string                     `json:"signatureV1"`
-	SignatureV2  string                     `json:"signatureV2"`
-	Data         interface{}                `json:"data"`
+	Validity     *time.Time                      `json:"validity"`
+	Sequence     uint64                          `json:"sequence"`
+	TTL          *uint64                         `json:"ttl"`
+	PublicKey    string                          `json:"publicKey"`
+	SignatureV1  string                          `json:"signatureV1"`
+	SignatureV2  string                          `json:"signatureV2"`
+	Data         interface{}                     `json:"data"`
 }
 
 func unmarshalIPNSEntry(data []byte) (*ipns_pb.IpnsEntry, error) {
@@ -90,7 +88,6 @@ func unmarshalIPNSRecord(entry *ipns_pb.IpnsEntry) (*IpnsInspectEntry, error) {
 		result.Validity = &validity
 	}
 
-
 	return &result, nil
 }
 
@@ -99,49 +96,3 @@ type IpnsInspectValidation struct {
 	Reason    string
 	PublicKey peer.ID
 }
-
-func verify(key string, entry *ipns_pb.IpnsEntry) (*IpnsInspectValidation, error) {
-	id, err := peer.Decode(key)
-	if err != nil {
-		return nil, err
-	}
-
-	validation := &IpnsInspectValidation{
-		PublicKey: id,
-	}
-
-	pub, err := id.ExtractPublicKey()
-	if err != nil {
-		// Make sure it works with all those RSA that cannot be embedded into the
-		// Peer ID.
-		if len(entry.PubKey) > 0 {
-			pub, err = ic.UnmarshalPublicKey(entry.PubKey)
-			if err != nil {
-				return nil, err
-			}
-
-			// Verify the public key matches the name we are verifying.
-			entryID, err := peer.IDFromPublicKey(pub)
-
-			if err != nil {
-				return nil, err
-			}
-
-			if id != entryID {
-				return nil, fmt.Errorf("record public key does not match the verified name")
-			}
-		}
-	}
-	if err != nil {
-		return nil, err
-	}
-
-	err = ipns.Validate(pub, entry)
-	if err == nil {
-		validation.Valid = true
-	} else {
-		validation.Reason = err.Error()
-	}
-
-	return validation, nil
-}
diff --git a/tooling/ipns/record.go b/tooling/ipns/record.go
index 49be37a..817db29 100644
--- a/tooling/ipns/record.go
+++ b/tooling/ipns/record.go
@@ -1,7 +1,9 @@
 package ipns
 
 import (
+	"github.com/ipfs/boxo/ipns"
 	ipns_pb "github.com/ipfs/boxo/ipns/pb"
+	"github.com/libp2p/go-libp2p/core/peer"
 )
 
 type IpnsRecord struct {
@@ -36,12 +38,11 @@ func (i *IpnsRecord) Key() string {
 	return i.Entry.PublicKey
 }
 
-func (i *IpnsRecord) Verify() (bool, error) {
-	result, err := verify(i.Entry.PublicKey, i.Pb)
-
+func (i *IpnsRecord) Valid() error {
+	id, err := peer.Decode(i.Entry.PublicKey)
 	if err != nil {
-		return false, err
+		return err
 	}
 
-	return result.Valid, nil
+	return ipns.ValidateWithPeerID(id, i.Pb)
 }
```

I would also remove the `IpnsInspectEntry` and rework it somehow else. As I mentioned in the Boxo PR, that struct was made specifically for the `ipns name inspect` command and I don't think it's adequate to be used here.

BigLep:
> Sadly, I can't commit here, so here it is:

@hacdias or @laurentsenta : lets add the w3dt-stewards team as an owner of this repo please.

laurentsenta:
(team added in https://github.com/ipfs/github-mgmt/pull/138 cc @BigLep @hacdias)

tooling/ipns/ipns.go
```diff
@@ -0,0 +1,59 @@
...
+	"path/filepath"
+	"regexp"
+
+	"github.com/ipfs/gateway-conformance/tooling/fixtures"
+)
+
+/**
+ * Extracts the public key from the path of an IPNS record.
+ * The path is expected to be in the format of:
+ * some/path/then/[pubkey](_anything)?.ipns-record
```
laurentsenta, line 16: Dear reviewer, one highlight for this PR:
we need to move the pubkey around with the ipns-record, we keep it in the filename, which gets the job done for now, but is not the greatest solution (ideally we'll store them in the car encoding)

laurentsenta:
Team, thanks for the feedbacks, and the patch @hacdias, the PR is complete and ready for review!

Note the `Test` won't pass until we release a new Kubo version, but `Test Master` is expected to pass.

> Very nice overall. I also wanted to ask if there is any reason why we're keeping the bash code from the old sharness tests as comments. I would personally like to get them cleanup up, but I just want to understand the reasoning. Is it to make it clear which tests are which during the transitioning period from sharness to the gateway conformance suite?

Exactly, many tests are not complete because we're missing fixtures (like ipns record here), so keeping the comments helps keep track of what's done, what's in progress, and what's not started. When we're done porting the tests, we'll remove the comment and mark this as our v1!

.github/workflows/test-master.yml
```diff
@@ -0,0 +1,70 @@
...
+              ipfs dag import "$car"
+          done
+
+          # Import ipns records
+          records=$(find ./fixtures -name '*.ipns-record')
+          for record in $records
+          do
+              key=$(basename -s .ipns-record "$record" | cut -d'_' -f1)
+              ipfs routing put --allow-offline "/ipns/$key" "$record"
+          done
```
laurentsenta, line 51: Reviewers, note this is how we move ipns records fixtures around,
we use a `pubkey_...ipns-record` file in the fixtures folder.

tests/t0116_gateway_cache_test.go
```diff
@@ -310,3 +220,103 @@ func TestGatewayCache(t *testing.T) {
 
 	Run(t, tests)
 }
+
+func TestGatewayCacheWithIPNS(t *testing.T) {
+	fixture := car.MustOpenUnixfsCar("t0116/gateway-cache.car")
+	ipns := ipns.MustOpenIPNSRecordWithKey("t0116/k51qzi5uqu5dlxdsdu5fpuu7h69wu4ohp32iwm9pdt9nq3y5rpn3ln9j12zfhe.ipns-record")
+	ipnsId := ipns.Key()
```
laurentsenta, line 227: highlight: this is the API to use an ipns record

tooling/ipns/record.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/ipfs/boxo/ipns"
+	ipns_pb "github.com/ipfs/boxo/ipns/pb"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+type IpnsRecord struct {
+	pb  *ipns_pb.IpnsEntry
+	key string
+	validity time.Time
+}
```
hacdias, line 15: Happy we didn't bring the `IpnsInspectEntry` here 😃

hacdias (approved):
Some minor notes, but everything else seems 👍

tests/t0116_gateway_cache_test.go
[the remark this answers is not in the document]
laurentsenta: Good point, fixed!

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,113 @@
...
+		// curl "http://127.0.0.1:$GWAY_PORT/ipns/$IPNS_KEY?format=ipns-record" > curl_output_filename &&
+		// ipfs name inspect --verify $IPNS_KEY < curl_output_filename > verify_output &&
+		// test_should_contain "$FILE_CID" verify_output
+		// '
+		// test_expect_success "GET KEY with format=ipns-record has expected HTTP headers" '
+		// curl -sD - "http://127.0.0.1:$GWAY_PORT/ipns/$IPNS_KEY?format=ipns-record" > curl_output_filename 2>&1 &&
+		// test_should_contain "Content-Disposition: attachment;" curl_output_filename &&
+		// test_should_contain "Content-Type: application/vnd.ipfs.ipns-record" curl_output_filename &&
+		// test_should_contain "Cache-Control: public, max-age=3155760000" curl_output_filename
+		// '
```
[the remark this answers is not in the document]
laurentsenta: For now, I'd like to keep them until we're done and we remove them all, it helps me keep track of what's done, what's in progress, and what's not started.

.github/workflows/test-master.yml
galargh, line 1: How about we add a matrix for Kubo versions we want to test with to the `test.yml` instead? Then we wouldn't need this one at all.

.github/workflows/test.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.1
+          go-version: 1.20.4
```
galargh, line 17: We don't actually use Go here right now, do we?

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,113 @@
...
+				Header("Accept", "application/vnd.ipfs.ipns-record"),
+			Response: Expect().
+				Headers(
+					Header("Content-Disposition").Contains("attachment;"),
+					Header("Content-Type").Contains("application/vnd.ipfs.ipns-record"),
+					Header("Cache-Control").Contains("public, max-age=3155760000"),
+				).
+				Body(
+					IsIPNSKey(ipnsKey).
+						IsValid().
```
galargh, line 88: `IsValid()` is obsolete, right?
```suggestion
					IsIPNSKey(ipnsKey)
```

tooling/ipns/check_test.go
```diff
@@ -0,0 +1,24 @@
...
+func TestIpnsCanOpenARecord(t *testing.T) {
+	path := "./_fixtures/k51qzi5uqu5dgh7y9l90nqs6tvnzcm9erbt8fhzg3fu79p5qt9zb2izvfu51ki.ipns-record"
+
+	// read file:
+	data, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	check := IsIPNSKey("k51qzi5uqu5dgh7y9l90nqs6tvnzcm9erbt8fhzg3fu79p5qt9zb2izvfu51ki").IsValid()
```
galargh, line 19: Same here:
```suggestion
	check := IsIPNSKey("k51qzi5uqu5dgh7y9l90nqs6tvnzcm9erbt8fhzg3fu79p5qt9zb2izvfu51ki")
```

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,113 @@
...
+				Query("format", "ipns-record"),
+			Response: Expect().
+				Headers(
+					Header("Content-Disposition").Contains("attachment;"),
+					Header("Content-Type").Contains("application/vnd.ipfs.ipns-record"),
+					Header("Cache-Control").Contains("public, max-age=3155760000"),
+				).
+				Body(
+					IsIPNSKey(ipnsKey).
+						IsValid().
```
galargh, line 60: And here:
```suggestion
					IsIPNSKey(ipnsKey)
```

tests/t0124_gateway_ipns_record_test.go
galargh: Alternatively, maybe this would be more straightforward
```suggestion
					IPNSKey(ipnsKey).
						IsValid().
```

tooling/ipns/ipns_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var YEAR_100 time.Time
+
+func init() {
+	YEAR_100 = time.Time(time.Date(2123, time.March, 01, 01, 0, 0, 801257000, time.UTC))
```
galargh, line 13: Should we check the current year and add 100?

tooling/test/test.go
```diff
@@ -126,6 +128,10 @@ func Run(t *testing.T, tests SugarTests) {
 
 			// send request
 			log.Debugf("Querying %s", url)
+			timeout, cancel := context.WithTimeout(context.Background(), 2 * time.Minute)
+			defer cancel()
+			req = req.WithContext(timeout)
```
galargh, line 133: Why is this necessary? IIRC there is a Go testcase timeout that one could configure to exit early. If we want to control it at this level too, I think it should be configurable.

galargh:
> Sadly, I can't commit here

Just an FYI, in a situation like this, you could create a fork and propose changes through a PR to the PR branch. And in case of smaller changes, you can always suggest them through a [comment](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request).

tests/t0124_gateway_ipns_record_test.go
laurentsenta: What makes you think it's obsolete?

I would keep it since it's one predicate we'd rather make explicit:

```
x := IsTarfile() // returns a Check function, calling x.Check(bytes) tells you whether a stream of bytes is a tar file,
y := IsTarFile().HasFile("file.txt") // A more precise Check functions

i := IsCarFile()
j := IsCarFile().HasBlock(....id)

m := IsIPNSKey(...)
n := IsIPNSKey(...).IsValid() // A more precise check function that verify the validity of the key
```

tooling/ipns/ipns_test.go
laurentsenta: Nope, we're testing a static fixture, the ipns key is valid in March 2023 + 100 years, that's what we unit-test.

tooling/test/test.go
laurentsenta: As far as I understand, there is only a single, global, go timeout: https://github.com/golang/go/issues/48157. That run hang'd on a single test for example: https://github.com/ipfs/gateway-conformance/actions/runs/4937755800/jobs/8826765779

> If we want to control it at this level too, I think it should be configurable.

I'd like to wait until we need that feature ([YAGNI](https://martinfowler.com/bliki/Yagni.html)). It's not hard to add but it'd make the code, configuration, and PR more complex without a use case right now.

.github/workflows/test.yml
galargh: Oh, but we will with the matrix!

laurentsenta: Thanks for the catch; probably a leftover from before the containerization 👍 

Note I'll merge the two tests.yml, which means we'll need go to call:

```yaml
- run: go install github.com/ipfs/kubo/cmd/ipfs@${{ matrix.target }}
```

In place of

```yaml
# only work with released binaries
- uses: ipfs/download-ipfs-distribution-action@v1
```

.github/workflows/test-master.yml
laurentsenta: That's much better, thanks for the idea

tests/t0124_gateway_ipns_record_test.go
galargh: https://github.com/ipfs/gateway-conformance/blob/16f9bba1698ac4fdab0ee12b4ac06512fe101e46/tooling/ipns/check.go#L17-L23

The validity check is always. The API doesn't allow you to skip it. I guess the questions are: do we ever wanna allow invalid IPNS keys and/or are invalid IPNS keys IPNS keys?

.github/workflows/test.yml
```diff
@@ -7,33 +7,51 @@ on:
 jobs:
   test:
     runs-on: 'ubuntu-latest'
+    strategy:
+      matrix:
+        target: ['latest', 'master']
```
galargh, line 12:
```suggestion
    strategy:
      fail-fast: false
      matrix:
        target: ['latest', 'master']
```

tests/t0124_gateway_ipns_record_test.go
laurentsenta: @lidel @hacdias do you have an opinion on this?

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,113 @@
...
+				Path("ipns/%s", ipnsKey).
+				Query("format", "ipns-record"),
+			Response: Expect().
+				Headers(
+					Header("Content-Disposition").Contains("attachment;"),
+					Header("Content-Type").Contains("application/vnd.ipfs.ipns-record"),
+					Header("Cache-Control").Contains("public, max-age=3155760000"),
+				).
+				Body(
+					IsIPNSKey(ipnsKey).
```
lidel, line 59: nit: what we get in response  body is not [IPNS Key](https://specs.ipfs.tech/ipns/ipns-record/#ipns-keys) :)   It is [IPNS Record](https://specs.ipfs.tech/ipns/ipns-record/#ipns-record).


- the only "key" is the Multihash of the public key used in  `/ipns/{libp2p-key}` aka `/ipns/{ipns-name}` -- https://specs.ipfs.tech/ipns/ipns-record/#ipns-name
  - we try to avoid "key" in other contexts, and use "ipns name" to avoid confusion 
- the body is[ IPNS Record](https://specs.ipfs.tech/ipns/ipns-record/#ipns-record) serialized as protobuf -- https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format

@laurentsenta mind renaming this helper to use `IPNSKey` → `IPNSRecord`?

README.md
```diff
@@ -111,6 +114,24 @@ The `extract-fixtures` command is used to extract the test fixtures from the `ga
...
+### Provisioning the Gateway
+
+We make minimal assumptions about the capabilities of the gateway being tested. Which means that we don't require nor expect the gateway to be writable. Therefore, you need to provision the gateway with test fixtures before running the test suite.
+
+These fixtures are located in the `./fixtures` folders. We distribute tools for extracting them. Refer to the documentation for the `extract-fixtures` command for more details.
+
+**Fixtures:**
+
+- Blocks & Dags: These are served as CAR file(s).
+- IPNS Records: These are distributed as files containing raw protocol buffers payloads. The file name includes the public key in this format: `pubkey(_optional_suffix)?.ipns-record`. We may decide to [share CAR files](https://github.com/ipfs/specs/issues/369) in the future.
```
lidel, line 132:
```suggestion
- IPNS Records: These are distributed as files containing [IPNS Record](https://specs.ipfs.tech/ipns/ipns-record/#ipns-record) [serialized as protobuf](https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format). The file name includes the Multihash of the public key ([IPNS Name](https://specs.ipfs.tech/ipns/ipns-record/#ipns-name)) in this format: `pubkey(_optional_suffix)?.ipns-record`. We may decide to [share CAR files](https://github.com/ipfs/specs/issues/369) in the future.
```

README.md
```diff
@@ -111,6 +114,24 @@ The `extract-fixtures` command is used to extract the test fixtures from the `ga
...
+
+### Provisioning the Gateway
+
+We make minimal assumptions about the capabilities of the gateway being tested. Which means that we don't require nor expect the gateway to be writable. Therefore, you need to provision the gateway with test fixtures before running the test suite.
+
+These fixtures are located in the `./fixtures` folders. We distribute tools for extracting them. Refer to the documentation for the `extract-fixtures` command for more details.
+
+**Fixtures:**
+
+- Blocks & Dags: These are served as CAR file(s).
```
lidel, line 131:
```suggestion
- Blocks & Dags: These are served as [CAR](https://ipld.io/specs/transport/car/) file(s).
```

tests/t0116_gateway_cache_test.go
```diff
@@ -310,3 +220,103 @@ func TestGatewayCache(t *testing.T) {
...
+func TestGatewayCacheWithIPNS(t *testing.T) {
+	fixture := car.MustOpenUnixfsCar("t0116/gateway-cache.car")
+	ipns := ipns.MustOpenIPNSRecordWithKey("t0116/k51qzi5uqu5dlxdsdu5fpuu7h69wu4ohp32iwm9pdt9nq3y5rpn3ln9j12zfhe.ipns-record")
+	ipnsKey := ipns.Key()
+
+	tests := SugarTests{
+		{
+			Name: "GET for /ipns/ unixfs dir listing succeeds",
+			Request: Request().
+				Path("ipns/%s/root2/root3/", ipnsKey),
```
lidel, line 233: unrelated nit:  every time we reference URL path, or IPFS content path, it should start with `/`, 
especially when we use `Equals("/ipns/%s/root2/root3/"` few lines below

lidel (changes requested):

tests/t0124_gateway_ipns_record_test.go
lidel: I think we want to be able to test negative scenarios like:
- error on bytes that are not a valid protobuf with ipns record ([spec](https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format)) (e.g. corrupted mid-flight)
-  error on invalid record  (e.g. valid protobuf, invalid signature)

So I would keep validation in a separate step (`IsValid` func).

btw: See my note on  renaming :upside_down_face:  it -- https://github.com/ipfs/gateway-conformance/pull/37#pullrequestreview-1429480277

```go
IsIPNSRecord(...) // calling x.Check(bytes) tells you whether a stream of bytes is an IPNS key that follows protobuf from https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format
IsIPNSRecord(...).IsValid() // A more precise check function that verify the validity of the key based on  checks described in https://specs.ipfs.tech/ipns/ipns-record/#record-verification
```

tooling/ipns/ipns.go
lidel: :+1:  I think this is a very sensible convention until we have https://github.com/ipfs/specs/issues/369 figured  out.

.github/workflows/test.yml
```diff
@@ -7,33 +7,52 @@ on:
...
+          do
+              ipfs dag import "$car"
+          done
+
+          # Import ipns records
+          records=$(find ./fixtures -name '*.ipns-record')
+          for record in $records
+          do
+              key=$(basename -s .ipns-record "$record" | cut -d'_' -f1)
+              ipfs routing put --allow-offline "/ipns/$key" "$record"
```
lidel, line 54: :thought_balloon:  (not a blocker, just a note for the future)  when https://github.com/ipfs/specs/issues/369 is figured out, we will be able to remove our reliance on `ipfs routing put`, which will simplify orchestration of Kubo.

tests/t0116_gateway_cache_test.go
laurentsenta: That's fair, I'll create a follow-up task + PR to fix this!

tests/t0124_gateway_ipns_record_test.go
laurentsenta: Fixed, thanks for the clarification!

tests/t0124_gateway_ipns_record_test.go
laurentsenta: @galargh I can see why that makes sense now (we have a `panic("not implemented")` on `false`), but that would create an opportunity for mistakes later: when we implement the `false` case, if we forgot about the default value, then the Check will accept almost any random stream of bytes.

I defaulted to nil, and require `IsValid()` or `IsInvalid()` to be called, let me know what you think.

galargh: Sounds good :)

tests/t0124_gateway_ipns_record_test.go
```diff
@@ -0,0 +1,113 @@
...
+		{
+			Name: "GET KEY with format=ipns-record has expected HTTP headers and valid key",
+			Request: Request().
+				Path("ipns/%s", ipnsName).
+				Query("format", "ipns-record"),
+			Response: Expect().
+				Headers(
+					Header("Content-Disposition").Contains("attachment;"),
+					Header("Content-Type").Contains("application/vnd.ipfs.ipns-record"),
+					Header("Cache-Control").Contains("public, max-age=3155760000"),
```
lidel, line 56: @laurentsenta all tests use the same timestamp, humans are lazy, which means some implementers will just hard-code `Cache-Control` and call it a day :P 

Mind adding one more test that has record with slightly different TTL, and call it out in the test Name  that Cache-Control  max-age MUST be based on the TTL from the IPNS Record? (fine to do it in a separate PR)

lidel (approved):
lgtm, thank you!

tests/t0124_gateway_ipns_record_test.go
laurentsenta: Will do, I created https://github.com/ipfs/gateway-conformance/issues/51

Issue #40, opened 2023-04-27
IPFS Things 2023 - Workshop Feedbacks


laurentsenta:
- [ ] docker build time is too slow
- [ ] window support!

String Formatting:

- [ ] Provide a non-formatted option for string comparison (`Equal vs Equalf`) to help with testing URL-escaped strings
- [ ] golang's verb collides with URL-escaping a lot: Provide a simpler templating system where you could create a string with something like `/ipfs/{{CID}}/%c4%85/%c4%99` and not have to deal with escaping strings 

Usage:

- [ ] Avoid magic - We should NOT default to "follow redirects"; make sure this is an explicit option
- [ ] Better description of fixtures

Fixtures:

- [ ] better description of fixtures, what it contains, and what's inside
- [ ] Make sure fixtures (CAR) can be recreated, maybe generate them on the fly
- [ ] provide a way to inspect these CARs

Debugging

- [ ] verbose mode: print content of fixtures, make them inspectable

Documentation

- [ ] document what are CAR files and why we use them

laurentsenta:
One open question we'll have to tackle: one of the feedback was "Make sure fixtures (CAR) can be recreated, maybe generate them on the fly".

Another of our goal is to have specs come with fixtures directly (like the [_redirect](https://github.com/ipfs/specs/pull/290/files) fixtures).

We'll have to think about:
- do we really want to generate fixtures on the fly? (one of our task was to NOT generate them on the fly in Kubo)
- what should the spec recipe, used to recreate it, look like? In kubo they look like [this](https://github.com/ipfs/kubo/tree/master/test/sharness/t0116-gateway-cache)

PR #41, opened 2023-05-01
feat: introduce templates

laurentsenta:
Contributes to https://github.com/ipfs/gateway-conformance/issues/40

>  golang's verb collides with URL-escaping a lot: Provide a simpler templating system where you could create a string with something like /ipfs/{{CID}}/%c4%85/%c4%99 and not have to deal with escaping strings

Introduce a templating system that lets you write:

```go
// named value "CID"
Fmt("/ipfs/{{CID}}/%c4%85/%c4%99", fixture.myCID)
// Values are replaced in the order they are defined, and you can reuse named values, and we recommend using backticks to avoid escaping strings.
Fmt(`<a href="{{cid}}">{{title}}}</a><a href="{{cid}}/index.html">index.html`, fixture.myCID, "Link Title!")
```

This templating is designed to be readable and quick to write which is why we don't reuse go template. The templating looks and behaves more like `sprintf`.

## Todo

- [x] complete implementation
- [x] test the "base string" case
- [x] implement escaping (so we CAN use `{{something}}` or `{{}}` as raw values when needed)
- [x] documentation
- [x] have the team review and ack the template


galargh (approved):

laurentsenta:
@galargh I implemented escaping and documentation, could you give it another review?

tests/t0116_gateway_cache_test.go
```diff
@@ -138,47 +138,47 @@ func TestGatewayCache(t *testing.T) {
 		{
 			Name: "GET for /ipfs/ file with matching Etag in If-None-Match returns 304 Not Modified",
 			Request: Request().
-				Path("ipfs/%s/root2/root3/root4/index.html", fixture.MustGetCid()).
+				Path("ipfs/{{}}/root2/root3/root4/index.html", fixture.MustGetCid()).
 				Headers(
-					Header("If-None-Match", fmt.Sprintf("\"%s\"", fixture.MustGetCid("root2", "root3", "root4", "index.html"))),
+					Header("If-None-Match", Templated("\"{{}}\"", fixture.MustGetCid("root2", "root3", "root4", "index.html"))),
```
lidel, line 143: This still introduces some golang-specific escaping.
I wonder if we could avoid escaping quotes by backticks?

```suggestion
					Header("If-None-Match", Verbatim(`"{{etag-of-index-html}}"`, fixture.MustGetCid("root2", "root3", "root4", "index.html"))),
```

lidel (changes requested):
This is really nice! 

I think we can make UX even better if:
- rename the public `Templated` to something shorter? I see we will use it directly in some tests to create path or header values. Maybe just `Fmt`? 
- we leverage backtick quotes to avoid escaping double quotes, and document it as best practice
- error on `{{}}`, always require meaningful name
  - i know its handy, but people are lazy and if we dont enforce name to be present they will use `{{}}`, and we want tests to be easy to read and reason about (avoid monsters like `{{}}://{{}}.ipfs.{{}}/ipfs/{{}}` or `"{{}},{{}},{{}},{{}},{{}}"`)

(see examples in suggestions inline, lmk how feasible these ideas are)

tests/t0116_gateway_cache_test.go
laurentsenta: Really cool, I switched every string with escaping to use backticks

laurentsenta:
Thanks for the reviews @lidel, @galargh 

- rename the public Templated to something else, preferably shorter? => Done
- leverage backtick quotes to avoid escaping double quotes => Done
- enforce names (avoid monsters like {{}}://{{}}.ipfs.{{}}/ipfs/{{}}) => Done!

galargh (approved):

README.md
```diff
@@ -151,6 +151,54 @@ gateway-conformance extract-fixtures
...
+It is required to always provide a meaningful `{{name}}`:
+
+```golang
+Fmt("/ipfs/{{cid}}/%c4%85/%c4%99", fixture.myCID) // => "/ipfs/Qm..../%c4%85/%c4%99"
+```
+
+Values are replaced in the order they are defined, and you may reuse named values
+
+```golang
+Fmt("<a href=\"{{cid}}\">{{label}}}</a><a href=\"{{cid}}/index.html\">index</a>", fixture.myCID, "Link Title!") // => '<a href="Qm...">Link Title!</a><a href="Qm..../index.html">index</a>'
```
lidel, line 184: nit:  avoid escoping `\"` in examples?

```suggestion
Fmt(`<a href="{{cid}}">{{label}}}</a><a href="{{cid}}/index.html">index</a>`, fixture.myCID, "Link Title!") // => '<a href="Qm...">Link Title!</a><a href="Qm..../index.html">index</a>'
```

README.md
```diff
@@ -151,6 +151,54 @@ gateway-conformance extract-fixtures
...
+```
+
+You may escape `{{}}` by using more than two opening or closing braces,
+
+```golang
+Templated("{}") // => "{}"
+Templated("{{{}}}") // => "{{}}"
+Templated("{{{{}}}}") // => "{{{}}}"
+Templated("{{{something}}}") // => {{something}}
```
lidel, line 193: i think empty is no longer allowed?

```suggestion
Fmt("{foo}") // => "{foo}"
Fmt("{{{foo}}}") // => "{{foo}}"
Fmt("{{{{foo}}}}") // => "{{{foo}}}"
Fmt("{{{foo}}}") // => {{foo}}
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- `Fmt` a string interpolation that replaces golang's and works better with HTML entities.
```
lidel, line 9:
```suggestion
- `Fmt` a string interpolation that replaces golang's and works better with HTML entities, and HTTP headers and URLs.
```

lidel (approved):
lgtm, small cosmetic questions for README below, but feel free to merge

PR #44, opened 2023-05-09
fix: return pointer in car check to be compatible with rest of API

laurentsenta:
Fixes the panic `body must be string, []byte, or a regular check` when we use a Car check.

- [x] Method on IsCar now return pointers instead of values which can be used in the `Body`
- [x] Add a test that relies on that API

### DIFF
--- a/tooling/check/car.go
+++ b/tooling/check/car.go
@@ -13,6 +13,8 @@ type CheckIsCarFile struct {
 	isOrdered bool
 }
 
+var _ Check[[]byte] = (*CheckIsCarFile)(nil)
+
 func IsCar() *CheckIsCarFile {
 	return &CheckIsCarFile{
 		blockCIDs: []cid.Cid{},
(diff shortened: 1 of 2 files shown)


tooling/check/car.go
```diff
@@ -30,38 +32,38 @@ func decoded(cidStr string) cid.Cid {
 	return cid
 }
 
-func (c CheckIsCarFile) HasBlock(cidStr string) CheckIsCarFile {
+func (c CheckIsCarFile) HasBlock(cidStr string) *CheckIsCarFile {
```
galargh, line 35: Do we want this too?
```suggestion
func (c *CheckIsCarFile) HasBlock(cidStr string) *CheckIsCarFile {
```

galargh (approved):

tooling/check/car.go
laurentsenta: (discussed out of band, we want to use values, and get clones)

PR #49, opened 2023-05-15
Add dag-scope tests

guanzo:
TODO:

* [x] Add a fixture that supports comprehensive car-scope tests.
* [x] Rename `car-scope` to `dag-scope` once lassie supports it.

How I generated the fixture
```sh
#!/bin/bash

# set kubo config "Internal.UnixFSShardingSizeThreshold": "10B"

mkdir -p dir/sub1
echo "abc123" > dir/sub1/hello.txt
mkdir -p dir/sub2
echo "xyz456" > dir/sub2/hello.txt

ROOT_CID=$(ipfs add -Qr --chunker=size-5 dir)

ipfs dag export ${ROOT_CID} > ./t0118-ipip-402-scope.car
```

The point of 2 subdirectories is that if the request is for one of them (`/sub1/hello.txt`), than the other (`/sub2`) should not appear in the CAR.

PR #56, opened 2023-05-19
feat: tests for IPIP-402 CAR parameters 

aschmahmann:
Conformance tests for IPIP-402.

- [x] Superseded and includes #49
- [ ] IPIP: https://github.com/ipfs/specs/pull/402
- [x] Boxo PR: https://github.com/ipfs/boxo/pull/303
- [x] Kubo PR: https://github.com/ipfs/kubo/pull/9914
- [x] **CI changes must be reverted before merging**: @hacdias updated this PR to run against the [Kubo PR](https://github.com/ipfs/kubo/pull/9914)

---

cc @guanzo @laurentsenta

Note: the last commits for cloning the requestbuilder and splitting out the repeatable tasks (e.g. format parameter + accept header testing) could use separate review on if they're a good idea/pattern.


tests/t0118_gateway_car_test.go
```diff
@@ -13,44 +14,179 @@ func TestGatewayCar(t *testing.T) {
...
+	formatParamReq := testWithFormatParam.Request
+	expected := testWithFormatParam.Response
+
+	carFormatQueryParams, found := formatParamReq.Query_["format"]
+	if !found {
+		t.Fatal("could not find 'format' query parameter")
+	}
+
+	if len(carFormatQueryParams) != 1 {
+		t.Fatal("only using a format parameter is supported")
```
laurentsenta, line 160:
```suggestion
		t.Fatal("only using a single format parameter is supported")
```

tests/t0118_gateway_car_test.go
```diff
@@ -13,44 +14,179 @@ func TestGatewayCar(t *testing.T) {
...
+				),
+		},
 	}
 
-	Run(t, tests)
+	transformedTests := standardCARTestTransforms(t, tests)
+	Run(t, transformedTests)
+}
+
+func standardCARTestTransforms(t *testing.T, tests SugarTests) SugarTests {
```
laurentsenta, line 123: That looks fine, I like the idea of using this decorator-ish approach,

We do call stuff like `Clone()` and `delete(map, k)` which moves away from our builder pattern and digs quite deep into the internals: could you move this into `tooling/helpers`?

tests/t0118_gateway_car_test.go
```diff
@@ -13,44 +14,179 @@ func TestGatewayCar(t *testing.T) {
...
+		},
+		{
+			Name: "GET CAR with dag-scope=entity",
+			Hint: `
+				dag-scope=entity should return a CAR file with all the blocks needed to 'cat'
+				a UnixFS file at the end of the specified path, or to 'ls' a UnixFS directory
+				at the end of the specified path.
+			`,
+			Request: Request().
+				Path("ipfs/%s", fixture.MustGetCid()).
```
laurentsenta, line 78:
```suggestion
				Path("ipfs/{{cid}}", fixture.MustGetCid()).
```

tests/t0118_gateway_car_test.go
```diff
@@ -13,44 +14,179 @@ func TestGatewayCar(t *testing.T) {
...
+				),
+		},
+		{
+			Name: "GET CAR with dag-scope=all",
+			Hint: `
+				dag-scope=all should return a CAR file with the entire contiguous DAG
+				that begins at the end of the path query, after blocks required to verify path segments.
+			`,
+			Request: Request().
+				Path("ipfs/%s/subdir", fixture.MustGetCid()).
```
laurentsenta, line 100:
```suggestion
				Path("ipfs/{{cid}}/subdir", fixture.MustGetCid()).
```

laurentsenta (changes requested):
really cool, thanks for the PR,

A note on moving file, and fixes, the `Master` test should pass, it's [broken because of templating right now.](https://github.com/ipfs/gateway-conformance/actions/runs/5047717583?pr=56)

laurentsenta:
Thanks for the update, I can see it's failing because of a few errors: https://github.com/ipfs/gateway-conformance/actions/runs/5060325462?pr=56#summary-13700442847

> Error: Body length mismatch: 2 != 3
> Error: Header 'Content-Disposition' expect to find substring 'attachment; filename="bafybeiefu3d7oytdumk5v7gn6s7whpornueaw7m7u46v2o6omsqcrhhkzi...
> Error: Header 'Content-Length' expected empty string, got '221' (CAR is streamed, gateway may not have the entire thing, unable to calculate total size)

The issue with Content-Length is a known bug with Kubo https://github.com/ipfs/kubo/issues/9651 we have a workaround, add the tests in https://github.com/ipfs/gateway-conformance/blob/c32f74d4d729b89b732e0943cf817abc1e1e3145/.github/workflows/test.yml#L65

Not sure about the two others, is this a new bug?

hacdias:
Notes:

- [ ] We have a lot of `.InThatOrder` - we may want to revisit that to check.

hacdias:
@lidel @aschmahmann tests are passing against `ipip-402` branch in Kubo, which uses the latest stuff in Boxo. I think we have a very good test coverage here.

tests/t0118_gateway_car_test.go
```diff
@@ -5,52 +5,596 @@ import (
...
+							subdirTwoSingleBlockFilesFixture.MustGetCid(),
+							subdirTwoSingleBlockFilesFixture.MustGetCid("subdir"),
+							subdirTwoSingleBlockFilesFixture.MustGetCid("subdir", "ascii.txt"),
+						).
+						Exactly().
+						InThatOrder(),
+				),
+		},
+		{
+			Name: "GET default CAR response of UnixFS file on a path with sharded directory",
```
lidel, line 43:
```suggestion
			Name: "GET default CAR response of UnixFS file on a path with HAMT-sharded directory",
```

tooling/car/unixfs.go
```diff
@@ -164,6 +186,62 @@ func (d *UnixfsDag) MustGetChildrenCids(names ...string) []string {
...
+func (d *UnixfsDag) MustGetCIDsInHAMTTraversal(path []string, child string) []string {
+	node := d.MustGetNode(path...)
+	var cids []string
+	tracker := dservTrackingWrapper{
+		DAGService: node.dsvc,
+	}
+	h, err := hamt.NewHamtFromDag(&tracker, node.node)
+	if err != nil {
+		panic(err)
+	}
```
lidel, line 234: I am concerned how future-proof this is, especially when we have no UnixFS specs yet.
We delegate "what HAMT is" to external library which may change at any time, changing the scope of accepted behaviors.

Let's [copy these explicit asserts](https://github.com/ipfs/boxo/blob/e2fc7f2fd0237afad200d7b0eec8b7a60bdc6644/ipld/unixfs/hamt/hamt.go#LL133C1-L149C3) here, so they are hardcoded here and executed before hamt library:

```suggestion
   pbnd, ok := node.node.(*dag.ProtoNode)
	if !ok {
		return nil, dag.ErrNotProtobuf
	}

	fsn, err := format.FSNodeFromBytes(pbnd.Data())
	if err != nil {
		return nil, err
	}

	if fsn.Type() != format.THAMTShard {
		return nil, fmt.Errorf("node was not a dir shard")
	}

	if fsn.HashType() != HashMurmur3 {
		return nil, fmt.Errorf("only murmur3 supported as hash function")
	}

	h, err := hamt.NewHamtFromDag(&tracker, node.node)
	if err != nil {
		panic(err)
	}
```

tests/t0118_gateway_car_test.go
```diff
@@ -5,52 +5,596 @@ import (
...
-					Exactly().
-					InThatOrder(),
-			),
+				Path("/ipfs/{{cid}}/subdir/ascii.txt", subdirTwoSingleBlockFilesFixture.MustGetCid()).
+				Query("format", "car"),
+			Response: Expect().
+				Status(200).
+				Body(
+					IsCar().
+						HasRoot(subdirTwoSingleBlockFilesFixture.MustGetCid()).
```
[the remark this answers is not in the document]
hacdias: I don't have very strong opinions, but considering the current behaviour of `ipfs dag import`, I think we can easily change the root of the CAR to be the last element in the path. That would allow for a better UX.

tests/t0118_gateway_car_test.go
```diff
@@ -5,52 +5,596 @@ import (
...
+			Name: "GET default CAR response with pathing through UnixFS Directory",
 			Hint: `
 				CAR stream is not deterministic, as blocks can arrive in random order,
 				but if we have a small file that fits into a single block, and export its CID
 				we will get a CAR that is a deterministic array of bytes.
 			`,
 			Request: Request().
-				Path("/ipfs/{{cid}}/subdir/ascii.txt", fixture.MustGetCid()).
-				Headers(
-					Header("Accept", "application/vnd.ipld.car"),
```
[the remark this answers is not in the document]
hacdias: All tests are run with `?format` and `Accept`. This is done via `StandardCARTestTransforms` which transforms the tests.

tooling/car/unixfs.go
hacdias: @lidel the problem with this is that we then have to copy more than the verification: we have to copy many parts from the `Shard` struct defined in the library, in order to perform the remaining operations. Are you sure? That'll likely be quite a bit of code.

tests/t0118_gateway_car_test.go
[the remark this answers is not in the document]
lidel: I've added 4 options to IPIP-402, let's continue there: https://github.com/ipfs/specs/pull/402#pullrequestreview-1465206498

tooling/car/unixfs.go
hacdias: I'll check

hacdias: [`9f9652f` (#56)](https://github.com/ipfs/gateway-conformance/pull/56/commits/9f9652fa334bf5f404861b4ca0208f3567712368)

tests/t0118_gateway_car_test.go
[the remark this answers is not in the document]
lidel: @hacdias @aschmahmann  based on feeback from Rod, I've relaxed requirement: https://github.com/ipfs/specs/pull/402#discussion_r1221517832
Means we need to update  conformance tests and expect either terminus CID (current behavior), or empty roots.

lidel (changes requested):

tests/t0118_gateway_car_test.go
```diff
@@ -5,52 +5,596 @@ import (
...
+				Body(
+					IsCar().
+						HasRoot(subdirTwoSingleBlockFilesFixture.MustGetCid()).
+						HasBlocks(
+							subdirTwoSingleBlockFilesFixture.MustGetCid(),
+							subdirTwoSingleBlockFilesFixture.MustGetCid("subdir"),
+							subdirTwoSingleBlockFilesFixture.MustGetCid("subdir", "ascii.txt"),
+						).
+						Exactly().
+						InThatOrder(),
```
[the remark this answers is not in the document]
hacdias: @lidel it's easy to just remove it if needed.

tests/t0118_gateway_car_test.go
```diff
@@ -5,52 +5,595 @@ import (
...
+		{
+			Name: "GET CAR with entity-bytes equivalent to a HTTP Range Request for the middle of a file",
+			Hint: `
+				The response MUST contain only the minimal set of blocks necessary for fulfilling the range request
+			`,
+			Request: Request().
+				Path("/ipfs/{{cid}}", subdirWithMixedBlockFiles.MustGetCidWithCodec(0x70, "subdir", "multiblock.txt")).
+				Query("format", "car").
+				Query("dag-scope", "entity").
+				Query("entity-bytes", "512:1023"),
```
lidel, line 469: :thought_balloon: this works for all-in-one implementations like boxo/gateway, but won't be enough for Saturn.  I don't want to block on this – filled https://github.com/ipfs/gateway-conformance/issues/75 as a follow-up work.

lidel (changes requested):
@aschmahmann  @hacdias fyi  I've filled https://github.com/ipfs/gateway-conformance/issues/75 to unblock landing this sooner than later.

iiuc the remaining asks are:

- [ ] update boxo code and assets for value in `CarV1Header.roots`, as noted in https://github.com/ipfs/specs/pull/402#discussion_r1221517832
- [ ] add regression test for `MustGetCIDsInHAMT*` helpers  (or fill issue for future PR)

I'm out Thu-Fri, but if above two are handled, lgtm to merge and ship.

hacdias:
This is blocked. Tests work well if the gateway we're testing against sends roots. If not, `car.NewReader` will scream at you saying there's no roots: https://github.com/ipfs/boxo/blob/aa12ba3e27965674d3e7cef7fefdccfa8d978d9c/ipld/car/car.go#L137-L139 (same in `ipld/go-car`, which I would prefer to use here).

The CAR specification also mentions the CAR requires 1 or more roots: https://ipld.io/specs/transport/car/carv1/#constraints

We have to reach a consensus on what to do about the CAR roots, then **update the libraries if we decide 0 roots is possible**. Release those libraries and then update here.

tooling/car/unixfs.go
```diff
@@ -164,6 +187,82 @@ func (d *UnixfsDag) MustGetChildrenCids(names ...string) []string {
 	return cids
 }
 
+// MustGetCidsInHAMT returns the cids in the HAMT at the given path. Does not include the CID of the HAMT root
+func (d *UnixfsDag) MustGetCidsInHAMT(names ...string) []string {
```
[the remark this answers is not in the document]
hacdias: Added. I used the same fixture as we have for the gateway. Note that I had to do some manual inspection for the known CIDs.

hacdias:
From standup: we decided to merge implementation in Boxo and Kubo, then this.

The decision on the CAR roots is deferred and both Boxo and conformance tests can be later updated.

⚠️ This tests currently check for 1 root (last segment) or no roots. The test logic is correct according tot he current IPIP-402 state. However, `go-car` does not support zero roots and therefore the tests will fail. Let's continue https://github.com/ipfs/specs/pull/402#discussion_r1222925940.

PR #65, opened 2023-06-01
fix: remove body check for subdomain redirection

hacdias:
See the context in https://github.com/ipfs/boxo/pull/326.


lidel (approved):
Lgtm, raitonale in https://github.com/ipfs/boxo/pull/326#issuecomment-1571078014, tldr this is Kubo-specific regacy hack, we should remove it from compliance suite.

PR #72, opened 2023-06-02
feat: body running in check

### LINKED ISSUE #67: Test the body in a golang `t.Run` structure
We have a system where each test for hears is registered as a golang test case:
https://github.com/ipfs/gateway-conformance/blob/17cbde1ec1968884336d5b2ceb415b63ccbfd62a/tooling/test/test.go#LL147C19-L147C19

It would be nice to have the same thing for the check on Body and Status Code
That would give us the granularity to temporarily disable a check like we do for Kubo's Content Length:

https://github.com/ipfs/kubo/blob/6eef0b4eefd6813dd6169d5a78ae57b9aa1cd90b/.github/workflows/gateway-conformance.yml#L77

Use case:
https://github.com/ipfs/gateway-conformance/pull/65

laurentsenta:
Closes #67

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - `Path(url)` does not add a leading `/` to the URL anymore.
 - Do not follow redirects by default anymore, remove `DoNotFollowRedirect` and add `FollowRedirect`.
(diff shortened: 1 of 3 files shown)


laurentsenta:
Failing test example: `/Body`:

https://github.com/ipfs/gateway-conformance/actions/runs/5153733330?pr=72

galargh (approved):

PR #79, opened 2023-06-16
feat: dedicated target per gateway specification and smaller-range groups

### LINKED ISSUE #74: Add dedicated compliance targets for Trustless Gateway
Discussed this briefly with @laurentsenta, projects like Saturn, Rhea, Boost etc will benefit from a test targets that only check compliance with https://specs.ipfs.tech/http-gateways/trustless-gateway/ spec.

We already have some prior art for [Subdomain Gateways and IPNS](https://github.com/ipfs/gateway-conformance/blob/81be5d516eba416c28586a2b33c3fda776cba704/tooling/specs/specs.go#L30-L32).

Ideally, we would provide targets for requested subsets of the spec (we know some use cases won't care about IPNS or CARs).

My thinking is to add:
1. `trustless-block-gateway` Just Blocks (no paths) – bare minimum, the most basic type of trustless gateway
2. `trustless-block-car-gateway`  Blocks and CARs with paths (@aarshkshah1992 would like to run this in Rhea projects)
3. `trustless-block-car-ipns-gateway` Blocks, CARs, and IPNS Records, enabling end-to-end verifiability of both `/ipfs` and `/ipns`. (we would run this in Kubo and bifrost-gateway)

cc @hacdias  – up to you if we add this to https://github.com/ipfs/gateway-conformance/pull/56 or do a follow-up PR

### LINKED ISSUE #78: Move `/ipfs/cid/sub?format=raw` to path gateway
https://specs.ipfs.tech/http-gateways/trustless-gateway/ does not allow paths for single blocks, but we want do support this on deserialized path gateway as it enables apps to fetch data via relative pathing (enables things like "autocodec" in HTML/JS+WASM).

Fix here would be to move `?format=raw` tests that use path out of https://github.com/ipfs/gateway-conformance/issues/74 and run only when path-gateway or subdomain-gateway is enabled.

hacdias:
Close #74. Close #78. This PR does the following:

- Defines the existing specs in https://specs.ipfs.tech/http-gateways/
- Some of the specs are decoupled in smaller sub-specifications. They can be enabled and disabled individually.
	- E.g. `+TrustlessGateway -TrustlessGatewayIPNS` will run CARs and RAW tests.
- Moves files around to make them more or less match the current specifications.
- Adds new tests for `/ipfs/{cid}?format=raw`
- @laurentsenta I kept in comment the name of the files were tests came from such that you can still go check if all tests were ported from sharness.
- Unexported `Run` such that **all** tests are attached to some specification.

What this PR **is not**:

- A perfect separation of tests per spec. There's some tests that are ambiguous and it's most likely because we always assumed "oh, it's Kubo and it can do everything". 
- Did not change the name of the fixture files.

Some **notes**:

- Most of the DNSLink Gateway (https://specs.ipfs.tech/http-gateways/dnslink-gateway) tests are not enabled yet.
- I want to make it clear that the DNSLink inlining in subdomain gateways (`{ dnslink-fqdn-inline }.ipns.{domain}`) is not a DNSLink gateway. 

---

I will also need some feedback here from @galargh @lidel @laurentsenta. Ideally, we should have all specifications defined and some (all?) would run by default. Then the user can tinker with it.

I think the most complicated part is defining subsets. For example: Trustless Gateway is a subset of the Path Gateway. Therefore, if I run the tests for the Path Gateway, I should also be running the Trustless Gateway tests, and so on and so forth.


hacdias:
@lidel @laurentsenta @galargh my idea for now would be something like this:

```go
const (
	// Real HTTP Gateway Specs: https://specs.ipfs.tech/http-gateways/
	PathGateway      Spec = "path-gateway"
	TrustlessGateway Spec = "trustless-gateway"
	DNSLinkGateway   Spec = "dnslink-gateway"
	SubdomainGateway Spec = "subdomain-gateway"
	RedirectsFile    Spec = "redirects-file" // This one can run for path gateway, dnslink and subdomain gateways.

	// TrustlessGateway sub-specifications.
	TrustlessBlockGateway Spec = "trustless-block-gateway"
	TrustlessCARGateway   Spec = "trustless-car-gateway"
	TrustlessIPNSGateway  Spec = "trustless-ipns-gateway"

	// Helper, smaller range specs.
	IPNSResolver    Spec = "ipns-resolver"
)
```

Enabling `TrustlessGateway` would enable `TrustlessBlockGateway`, `TrustlessCARGateway` and `TrustlessIPNSGateway`. But they could be also individually removed or enabled.

So my question here, and the food for thought we need is:

- Do we want to have groups of specs? If so, how will we handle it. Or do we prefer to simply have sub-things that run when simultaneously turned on, e.g., `PathGateway+RedirectsFile`
- All tests will be assigned a spec, so we should run all by default?

tooling/specs/specs.go
```diff
@@ -56,11 +82,23 @@ func (s Spec) IsEnabled() bool {
 }
 
 func (s Spec) Enable() {
-	specEnabled[s] = true
+	if specs, ok := specGroups[s]; ok {
+		for _, s := range specs {
+			specEnabled[s] = true
```
laurentsenta, line 87: You might need a recursive call here since we don't differentiate between leaves and groups: typing allows for dags of specs.

It might be nice to have a type for leaves and one for leaves+groups so we can enforce my point below regarding internal/external APIs.

tooling/specs/specs.go
```diff
@@ -27,16 +27,42 @@ func (m maturity) isMature() bool {
...
-	DNSLinkResolver:  stable,
-	IPNSResolver:     stable,
+	RedirectsFile:    stable,
+
+	IPNSResolver: stable,
+}
+
+// Groups of specifications that, if enabled or disabled, will enable or disable
+// all the respective specifications. The group specifications must not be used for
+// RunIfSpecsAreEnabled, only the sub-specifications.
```
laurentsenta, line 59: 👍

hacdias: Maybe as you said before, we can make some distinction between Leafs and Groups in order to ensure that is forced.

hacdias:
The PR is ready to looked at. Note that I updated the PR description with all the new information. It seems that the tests are stuck "provisioning Kubo gateway" and I'm not sure what's wrong as I did not change any of the fixtures.

laurentsenta (approved):
Code LGTM, investigating why CI breaks.

lidel (changes requested):
Thank you @hacdias, lgtm!

I've dropped some unrelated TODOs for future PRs (things I've noticed while looking at this PR), but as for this PR,  it's only missing docs (outline of mvp readme below)

> Ideally, we should have all specifications defined and some (all?) would run by default. Then the user can tinker with it.

Yes, by default we run all, and then user can adjust, if required.

I think this PR should also include section in README about this. 
Questions we want to have answered there:
- what is run by default (all mature tests)
- opt-outs
  - how user can skip specific test group (example:  disable subdomain and dnslink gateways)
  - how user can skip individual test  (example: `-skip` is used in Kubo CI [here](https://github.com/ipfs/kubo/blob/master/.github/workflows/gateway-conformance.yml#L97) but not documented anywhere in readme atm)
- how user can disable all tests and run only specific group
  - including example how to run only the trustless raw+car but without ipns will be enough

PR #84, opened 2023-06-22
feat: explicitly not check CAR roots

### LINKED ISSUE #83: IPIP-402: remove CAR roots checks from trustless gw tests
As noted in https://github.com/ipfs/specs/pull/402#issuecomment-1599428849 we were unable to get consensus on what goes into that field.

Updated spec states clients MAY ignore it, so the conformance tests should no longer make any assumptions about roots present in CAR header – we should remove checks for `HasRoot+MightHaveNoRoots` for now  (but keep these helpers, we may reintroduce them in the future).

I know its frustrating, as the time was spent on  adding `HasRoot` and `MightHaveNoRoots`  helpers for testing the previous iteration of IPIP-402, but this is the only path forward, given the cards we were dealt with.

hacdias:
Closes #83.

I had to add a `IgnoreRoots` option because the way the current CAR checks work, we **always** have to say something about the roots if we do a `Exactly`. So now you can ignore CAR roots on Exactly. Nevertheless, the logic for `CheckIsCarFile.Check` is getting quite hairy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- CAR tests no longer check for the roots. See discussion in [IPIP-402](https://github.com/ipfs/specs/pull/402).
+
 ## [0.1.0] - 2023-06-08
 ### Added
 - `Fmt` a string interpolation that replaces golang's and works better with HTML entities, and HTTP headers and URLs.
--- a/tests/trustless_gateway_car_test.go
+++ b/tests/trustless_gateway_car_test.go
@@ -30,8 +30,6 @@ func TestTrustlessCarPathing(t *testing.T) {
 				Status(200).
 				Body(
 					IsCar().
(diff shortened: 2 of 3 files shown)


laurentsenta:
What about we rewrite that check so we can write:

```go
					IsCar().
						ShouldHaveRoot(subdirWithMixedBlockFiles.MustGetCid("subdir", "multiblock.txt")).
						MustHaveBlock(
							subdirWithMixedBlockFiles.MustGetCid("subdir", "multiblock.txt"),
						MustHaveBlocks(
							subdirWithMixedBlockFiles.MustGetChildrenCids("subdir", "multiblock.txt")[0]),
						).
						Exactly().
						InThatOrder(),
```

?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- CAR tests no longer check for the roots. See discussion in [IPIP-402](https://github.com/ipfs/specs/pull/402) for more information.
```
laurentsenta, line 9:
```suggestion
### Added
- CAR tests no longer check for the roots. See discussion in [IPIP-402](https://github.com/ipfs/specs/pull/402).
```

PR #85, opened 2023-06-22
feat: issue-75 - test entity-bytes with missing blocks

### LINKED ISSUE #75: Add backend timeout test for entity-bytes from IPIP-402
https://github.com/ipfs/gateway-conformance/pull/56 introduced request-response tests for [IPIP-402](https://github.com/ipfs/specs/pull/402), including `entity-bytes` for range requests that return a minimal subset of blocks for passed byte range:

```go
Request().
Path("/ipfs/{{cid}}", subdirWithMixedBlockFiles.MustGetCidWithCodec(0x70, "subdir", "multiblock.txt")).
Query("format", "car").
Query("dag-scope", "entity").
Query("entity-bytes", "512:1023"),
```

Tests included in #56  will confirm the response includes the correct subset of blocks for the file.

### Extra mile for testing `entity-bytes` 

Because fixture includes all blocks available for  `multiblock.txt` file, the compliance test will not detect when a backend is broken and  downloads the entire thing from the beginning (`0:1023`) just to return a slice of that (`512:1023`). 

Such implementation will produce a valid response for cached data, but cache misses will be excruciatingly slow, and usually timeout.
If the file is 4GiB video and the client requested 1MiB  in the middle, nothing will happen for a long long time, because the backend will be busy retrieving 2GiB of data just to return the last 1MiB. 

I believe conformance suite should be testing this, because if our internal teams ended up with this footgun in Rhea/Saturn (brainstorming fix in https://github.com/filecoin-saturn/L1-node/issues/415), we most likely will see more of this  in the future.

### How to test this

We should have additional test with `partially-resolvable-1GiB-multiblock.bin`, where only root block + child block responsible for requested range are present in the CAR fixture.  

This way, the test will hang/timeout when an incorrectly implemented backend tries to fetch blocks that are not related to the `entity-bytes` request.

laurentsenta:
Closes #75 

- [x] rename fixtures/t0118 to match the test name
- [x] `MustGetChildren` -> `MustGetDescendants`
- [x] Implement "real" `MustListChildren`
- [x] Generate a car file with missing blocks
- [x] Add two tests that try to query the beginning and end of a car file with missing blocks
- [x] Add a `README.md` with a recipe for the fixtures.


tests/trustless_gateway_car_test.go
```diff
@@ -387,10 +387,80 @@ func TestTrustlessCarDagScopeAll(t *testing.T) {
...
+						MightHaveNoRoots().
+						HasBlocks(
+							missingBlockFixture.MustGetCid(),
+							missingBlockFixture.MustGetChildrenCids()[2],
+						).
+						Exactly(),
+				),
+		},
+		{
+			Name: "GET CAR with entity-bytes when missing a block will timeout",
```
lidel: @laurentsenta (just FYI) I've removed this one, as its body was the same as the first one we had – assuming it was just left-over of some copy-paste.

I did not fix it to actually test timeout, because:
- we don't want to introduce unnecessary latency to the test suite
- the duration after which returning HTTP 504  is not currently specified, so we can't test conformance anyway
- two existing tests are more than enough - especially `entity-bytes=2200:*` one – this alone will fail if implementation is broken the way I described in https://github.com/ipfs/gateway-conformance/issues/75

tooling/car/fixture_test.go
```diff
@@ -93,9 +93,54 @@ func TestIssue54MustGetChildren(t *testing.T) {
...
-	cids = f.MustGetChildrenCids("sub1")
+	cids = f.MustGetDescendantsCids("sub1")
 	assert.Len(t, cids, 3)
 	assert.Equal(t, "QmZgfvZtoFdbJy4JmpPHc1NCXyA7Snim2L8e6zKspiUzhu", cids[0])
 	assert.Equal(t, "QmaATBg1yioWhYHhoA8XSUqD1Ya91KKCibWVD4USQXwaVZ", cids[1])
 	assert.Equal(t, "QmdQEnYhrhgFKPCq5eKc7xb1k7rKyb3fGMitUPKvFAscVK", cids[2])
 }
+
+func TestMustGetChildrenDespiteMissingBlocks(t *testing.T) {
+	f := MustOpenUnixfsCar("./_fixtures/file-3k-and-3-blocks-missing-block.car")
```
lidel, line 104: :thought_balloon:  excellent call to have a separate fixture for testing tooling, independent of fixture used in actual gateway test :ok_hand:

tests/trustless_gateway_car_test.go
```diff
@@ -374,10 +374,60 @@ func TestTrustlessCarDagScopeAll(t *testing.T) {
...
+				dag-scope=entity&entity-bytes=y:* should return a CAR file with
+				only the blocks needed to fullfill the request. This MUST
+				succeed despite the fact that a block with bytes before 'y' is
+				not retrievable.
+			`,
+			Request: Request().
+				Path("/ipfs/{{cid}}", missingBlockFixture.MustGetCidWithCodec(0x70)).
+				Query("format", "car").
+				Query("dag-scope", "entity").
+				Query("entity-bytes", "2200:*"),
```
lidel, line 418: :information_source:  I had gut feeling to check what we send on the wire – we may need to add `QueryRaw` here, but not in scope of this PR, not a blocker –  filled separate https://github.com/ipfs/gateway-conformance/issues/88

lidel (approved):
Thank you @laurentsenta, this is exactly what we needed!

(pushed some minor tweaks to Name and Hint + some comments inline fyi,  but ok to merge as-is – we need this and to unblock https://github.com/filecoin-saturn/L1-node/issues/447)

tests/trustless_gateway_car_test.go
laurentsenta: Thanks for the update and removing that useless test 🙏 
I created an issue to add timeouts https://github.com/ipfs/gateway-conformance/issues/89

hacdias (approved):

galargh (approved):

PR #87, opened 2023-06-23
feat: CAR order and dups parameters from IPIP-412

hacdias:
This PR creates tests for [IPIP-412](https://github.com/ipfs/specs/pull/412). Note the following:

- ~⚠️ This PR is running against [`ipip-412`](https://github.com/ipfs/kubo/pull/9989) branch in Kubo. Disable this before merging.~
- Tests for duplicates use a fixture where a directory contains two files that are the same. If `dups=n`, then there are no duplicates. If `dups=y`, then the blocks of the file are sent twice, by the order they show up in the DAG. This is tested with `order=dfs`.
- The test above also tests  `order=dfs`.
- I added a test for `order=unk` where I simply check if all the CIDs are present, but _not exactly_ since the order can be unknown. This test also doesn't check for the order in the response content type **because** the gateway can answer with any order they want.


tests/trustless_gateway_car_test.go
```diff
@@ -579,6 +579,107 @@ func TestTrustlessCarEntityBytes(t *testing.T) {
...
+							dirWithDuplicateFiles.MustGetCid("hello.txt"),
+							dirWithDuplicateFiles.MustGetCid("multiblock.txt"),
+							dirWithDuplicateFiles.MustGetChildrenCids("multiblock.txt"),
+						)...),
+				),
+		},
+	}
+
+	// TODO: add sub-specification for these tests.
+	RunWithSpecs(t, tests, specs.TrustlessGatewayCAR)
```
hacdias, line 680: I want to add a specific one here. What do you think about `TrustlessGatewayAcceptCAR` for accept parameters?

laurentsenta: Is this anchored with something in the specs? Could we use the IPIP id instead?

tests/trustless_gateway_car_test.go
```diff
@@ -579,6 +579,107 @@ func TestTrustlessCarEntityBytes(t *testing.T) {
...
+					IsCar().
+						IgnoreRoots().
+						HasBlocks(flattenStrings(t,
+							dirWithDuplicateFiles.MustGetCid(),
+							dirWithDuplicateFiles.MustGetCid("ascii.txt"),
+							dirWithDuplicateFiles.MustGetCid("ascii-copy.txt"),
+							dirWithDuplicateFiles.MustGetCid("hello.txt"),
+							dirWithDuplicateFiles.MustGetCid("multiblock.txt"),
+							dirWithDuplicateFiles.MustGetChildrenCids("multiblock.txt"),
+						)...),
```
laurentsenta, line 674: golang's splicing is quite restricted :/ I'd like to get rid of `flattenStrings`, which is too low level for these test definitions; what about we use:

```suggestion
						HasBlock(dirWithDuplicateFiles.MustGetCid()).
						HasBlock(dirWithDuplicateFiles.MustGetCid("ascii.txt")).
						HasBlock(dirWithDuplicateFiles.MustGetCid("ascii-copy.txt")).
						HasBlock(dirWithDuplicateFiles.MustGetCid("hello.txt")).
						HasBlock(dirWithDuplicateFiles.MustGetCid("multiblock.txt")).
						HasBlocks(dirWithDuplicateFiles.MustGetChildrenCids("multiblock.txt")).
```

?

tests/trustless_gateway_car_test.go
```diff
@@ -579,6 +579,107 @@ func TestTrustlessCarEntityBytes(t *testing.T) {
...
+				Body(
+					IsCar().
+						IgnoreRoots().
+						HasBlocks(flattenStrings(t,
+							dirWithDuplicateFiles.MustGetCid(),
+							dirWithDuplicateFiles.MustGetCid("ascii.txt"), // ascii.txt = ascii-copy.txt
+							dirWithDuplicateFiles.MustGetCid("hello.txt"),
+							dirWithDuplicateFiles.MustGetCid("multiblock.txt"),
+							dirWithDuplicateFiles.MustGetChildrenCids("multiblock.txt"),
+						)...).
```
laurentsenta, line 644: Same comment as below, you might also be able to use something like:

```
X
    .HasBlocks(call1(), call2(), call3())
    .HasBlocks(callMany()...)
```

tests/trustless_gateway_car_test.go
```diff
@@ -579,6 +579,107 @@ func TestTrustlessCarEntityBytes(t *testing.T) {
 	RunWithSpecs(t, helpers.StandardCARTestTransforms(t, tests), specs.TrustlessGatewayCAR)
 }
 
+func TestTrustlessCarOrderAndDuplicates(t *testing.T) {
+	dirWithDuplicateFiles := car.MustOpenUnixfsCar("t0118/dir-with-duplicate-files.car")
```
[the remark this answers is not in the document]
laurentsenta: Also, @hacdias could you provide a recipe on how to create the fixture?

I added an example in https://github.com/ipfs/gateway-conformance/blob/main/fixtures/trustless_gateway_car/README.md#recipes,

lidel (changes requested):

tests/trustless_gateway_car_test.go
lidel: Maybe more generic `TrustlessGatewayCAROptional`? We may have more optional things.
This way we have a convention where people can do `-*optional`  and only run tests for mandatory things.

tests/trustless_gateway_car_test.go
```diff
@@ -631,6 +631,13 @@ func TestTrustlessCarEntityBytes(t *testing.T) {
 
 func TestTrustlessCarOrderAndDuplicates(t *testing.T) {
 	dirWithDuplicateFiles := car.MustOpenUnixfsCar("trustless_gateway_car/dir-with-duplicate-files.car")
+	multiblockCIDs := []string{
+		"bafkreie5noke3mb7hqxukzcy73nl23k6lxszxi5w3dtmuwz62wnvkpsscm",
+		"bafkreih4ephajybraj6wnxsbwjwa77fukurtpl7oj7t7pfq545duhot7cq",
+		"bafkreigu7buvm3cfunb35766dn7tmqyh2um62zcio63en2btvxuybgcpue",
+		"bafkreicll3huefkc3qnrzeony7zcfo7cr3nbx64hnxrqzsixpceg332fhe",
+		"bafkreifst3pqztuvj57lycamoi7z34b4emf7gawxs74nwrc2c7jncmpaqm",
+	}
```
laurentsenta, line 640: - Could you provide a recipe for the car file (in a README.md, there should be other examples)
- There is no way to avoid hardcoding these?

hacdias: @laurentsenta re: hardcoding https://github.com/ipfs/gateway-conformance/pull/87#discussion_r1253723076

tests/trustless_gateway_car_test.go
```diff
@@ -629,6 +629,168 @@ func TestTrustlessCarEntityBytes(t *testing.T) {
 	RunWithSpecs(t, helpers.StandardCARTestTransforms(t, tests), specs.TrustlessGatewayCAR)
 }
 
+func TestTrustlessCarOrderAndDuplicates(t *testing.T) {
+	dirWithDuplicateFiles := car.MustOpenUnixfsCar("trustless_gateway_car/dir-with-duplicate-files.car")
+	multiblockCIDs := []string{
```
laurentsenta, line 634:
```suggestion
        // This array is defined at the SPEC level and should not depend on library behavior
        // See the recipe for `dir-with-duplicate-files.car`
	multiblockCIDs := []string{
```

tests/trustless_gateway_car_test.go
```diff
@@ -398,7 +398,7 @@ func TestTrustlessCarEntityBytes(t *testing.T) {
 						IgnoreRoots().
 						HasBlocks(
 							missingBlockFixture.MustGetCid(),
-							missingBlockFixture.MustGetChildrenCids()[0],
+							"QmPKt7ptM2ZYSGPUc8PmPT2VBkLDK3iqpG9TBJY7PCE9rF",
```
laurentsenta, line 401:
```suggestion
                                                        // This CID is defined at the SPEC level
                                                        // See the recipe for `file-3k-and-3-blocks-missing-block.car`
							"QmPKt7ptM2ZYSGPUc8PmPT2VBkLDK3iqpG9TBJY7PCE9rF",
```

tests/trustless_gateway_car_test.go
```diff
@@ -423,7 +423,7 @@ func TestTrustlessCarEntityBytes(t *testing.T) {
 						IgnoreRoots().
 						HasBlocks(
 							missingBlockFixture.MustGetCid(),
-							missingBlockFixture.MustGetChildrenCids()[2],
+							"QmWXY482zQdwecnfBsj78poUUuPXvyw2JAFAEMw4tzTavV",
```
laurentsenta, line 426:
```suggestion
                                                        // This CID is defined at the SPEC level
                                                        // See the recipe for `file-3k-and-3-blocks-missing-block.car`
							"QmWXY482zQdwecnfBsj78poUUuPXvyw2JAFAEMw4tzTavV",
```

laurentsenta (approved):
lgtm, let's not hardcode CIDs in the code without leaving notes to future maintainers!

PR #96, opened 2023-06-28
feat: added support for modifying request/response before/after making the request

galargh:
This PR adds support for modifying request/response before/after making the request. This enables porting t0116 test cases that depend on reading the value of the Etag header and using it in subsequent requests.


galargh:
See https://github.com/ipfs/gateway-conformance/pull/97

PR #97, opened 2023-06-28
t0116: implement diretag checks

### LINKED ISSUE #93: port t0116
https://github.com/ipfs/gateway-conformance/pull/30

laurentsenta:
Closes #93

Rationale:

The API in https://github.com/ipfs/gateway-conformance/pull/96 relies on having implicit coupling (side effects and time) but the rest of the API is declarative.

This is a demo of how we could do the same without changing the API.

I would favor that approach:
- we don't endorse side-effects and transforming requests at the API level, we'd have a single "rogue" test,
- the side-effects are simple and explicit (we call `Run` twice, and we create requests in the regular way)
- no API change


tests/path_gateway_unixfs_test.go
```diff
@@ -286,32 +286,55 @@ func TestGatewayCache(t *testing.T) {
...
+		{
+			Name: "DirIndex etag is based on xxhash(./assets/dir-index-html), so we need to fetch it dynamically",
+			Request: Request().
+				Path("/ipfs/{{cid}}/root2/root3/", fixture.MustGetCid()),
+			Response: Expect().
+				Status(200).
+				Headers(
+					Header("Etag").
+						Checks(func(v string) bool {
+							etag = v[1 : len(v)-1] // trim quotes
```
galargh, line 306:
```suggestion
							etag = strings.Trim(v, `"`)
```

galargh (approved):

galargh:
👏

PR #103, opened 2023-06-30
feat: improve testing

laurentsenta:
- [x] fix tooling test suite
- [x] run go test tooling
- [x] more explicit workflow names

### DIFF
--- a/.github/workflows/test-kubo-e2e.yml
+++ b/.github/workflows/test-kubo-e2e.yml
@@ -1,4 +1,4 @@
-name: Test
+name: Test Kubo (e2e)
 
 on:
   workflow_dispatch:
--- a/.github/workflows/test-tooling.yml
+++ b/.github/workflows/test-tooling.yml
@@ -1,0 +1,31 @@
+name: Test Tooling
+
+on:
+  workflow_dispatch:
+  push:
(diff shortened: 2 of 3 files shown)


laurentsenta:
@galargh I'm thinking about adding unified CI to the repo, but I don't want to run the test in `./tests`... that's the project's output, they need provisioning, running daemons, etc. more like an e2e test.

I'm thinking either
1. we add build tags support, similar to `skip32bit` in [configuration ](https://github.com/protocol/.github#configuration-files),
2. we just remove the `./tests` folder during the [setup actions](https://github.com/protocol/.github#setup-actions).

(1) might not be useful outside of gway conformance and (2) seems to just work ¯\_(ツ)_/¯, wdyt?

galargh:
That's a swell idea! I think we should go with 2. (at least for now). The main reason is quite selfish, we need to do a looot of changes in uCI in July to get us closer to https://github.com/protocol/.github/issues/514 before Go 1.21 is out so if we can limit other changes in the meantime, that's great. Also, if we have uCI in Gateway Conformance, it'll be extremely useful for testing if our changes to uCI as a whole work as expected :)

galargh (approved):

Issue #104, opened 2023-06-30
Test production gateways


laurentsenta:
Contributes to https://github.com/ipfs/gateway-conformance/issues/8

- [x] pick one or more gateways (https://ipfs.github.io/public-gateway-checker/ - ipfs.runfission.com might be a good candidate)
- [x] automatically pin our fixtures.car file (using web3.storage) on merge to `main`
- [x] configure the test suite to run tests there (probably not the one related to dnslink and ipns)
- [x] generate dashboards

laurentsenta:
The follow-up task would be to create a dashboard when we run these tests:

1. run the tests for every gateway we care about
2. aggregate test results by specs involved (`gateway-car`, `gateway-trustless`, etc).
    ideally we end up with a matrix of `(gateway x specs x PASS/FAIL)`
3. generate an HTML, similar to https://ipfs.github.io/public-gateway-checker/

BigLep:
@laurentsenta : cool.  I can see how this can be done for gateways with a public URL like the ones in https://github.com/ipfs/public-gateway-checker .  How would we do this against various gateway implementations (Boxo, Kubo, bifrost-gateway)?  Would we be needing to recreate their setups in the gateway-conformance pipeline so we can get an instance of them running?

laurentsenta:
@BigLep 

> Would we be needing to recreate their setups in the gateway-conformance pipeline so we can get an instance of them running?

That would be a "worst case" scenario: since all the projects you listed have gateway conformance in their CI, we can definitely find a collaborative approach where we reuse their workflows or the artifacts they produced.

- https://github.com/ipfs/kubo/actions/workflows/gateway-conformance.yml
- https://github.com/ipfs/bifrost-gateway/actions/workflows/gateway-conformance.yml
- https://github.com/ipfs/boxo/actions/workflows/gateway-conformance.yml
- https://github.com/filecoin-saturn/L1-node/actions/workflows/gateway-conformance.yml

laurentsenta:
@lidel @BigLep 
fwiw I drafted a solution in https://github.com/ipfs/gateway-conformance/pull/116

example dashboard output: https://github.com/singulargarden/gateway-conformance/actions/runs/5681118160/attempts/1#summary-15396709726

laurentsenta:
Closing this, it's running. Follow-up / related tasks live in https://github.com/ipfs/gateway-conformance/issues/120

PR #108, opened 2023-07-04
feat: test production gateways

laurentsenta:
Contributes to #104

Add tooling to run tests over multiple gateways & merge the outputs into a single dashboard. This is using a few arbitrary production gateways for now.

- [x] merges fixtures are now a single car v1 with a single root (makes it compatible with web3.storage)
- [x] add test-prod-e2e CI test
    - [x] upload fixtures to web3.storage
    - [x] run gateway conformance test suite on a few (arbitrary) gateways
    - [x] aggregate results & generate dashboard
- [x] generate a badge
- [x] patch add-to-web3 https://github.com/web3-storage/add-to-web3/pull/86

Example of output: https://github.com/singulargarden/gateway-conformance/actions/runs/5680836848
Example of markdown: https://github.com/laurentsenta/gateway-conformance/blob/main/README.md (it links to ipfs/gateway-conformance, hence the gray status for now)

## follow-up

- Implement a similar test for dev gateways (see https://github.com/ipfs/gateway-conformance/issues/104#issuecomment-1621944117)
- tweak the production gateways we want to test and the specs we cover

### DIFF
--- a/.github/workflows/test-prod-e2e.yml
+++ b/.github/workflows/test-prod-e2e.yml
@@ -48,14 +48,21 @@ jobs:
           set -o pipefail
 
           pin_status=""
+          start_time=$(date +%s)
+          timeout=$((2 * 60)) # Timeout set to 2 minutes
 
           while [[ "$pin_status" != "Pinned" ]]; do
             response=$(curl --fail --oauth2-bearer "$W3STORAGE_TOKEN" "https://api.web3.storage/status/$CID")
             pin_status=$(echo $response | jq -r '.pins[0].status')
             echo "Current pin status: $pin_status"
 
-            if [[ "$pin_status" != "Pinned" ]]; then
+            current_time=$(date +%s)
+            elapsed_time=$((current_time - start_time))
+
+            if [[ "$pin_status" != "Pinned" && $elapsed_time -lt $timeout ]]; then
               sleep 10 # Sleep for 10 seconds before trying again
+            else
+              break
             fi
           done
   test:
--- a/.github/workflows/test-prod-e2e.yml
+++ b/.github/workflows/test-prod-e2e.yml
@@ -68,7 +68,6 @@ jobs:
   test:
     runs-on: "ubuntu-latest"
     strategy:
-      fail-fast: false
       matrix:
         target: ["ipfs.runfission.com", "w3s.link"]
     defaults:
@@ -110,8 +109,8 @@ jobs:
   aggregate:
     runs-on: "ubuntu-latest"
     needs: [test]
-    strategy:
-      fail-fast: true
+    # the tests might have failed
+    if: always()
     defaults:
       run:
         shell: bash
--- a/.github/workflows/test-prod-e2e.yml
+++ b/.github/workflows/test-prod-e2e.yml
@@ -50,7 +50,7 @@ jobs:
           pin_status=""
 
           while [[ "$pin_status" != "Pinned" ]]; do
-            response=$(curl --oauth2-bearer "$W3STORAGE_TOKEN" "https://api.web3.storage/status/$cid")
+            response=$(curl --oauth2-bearer "$W3STORAGE_TOKEN" "https://api.web3.storage/status/$CID")
             pin_status=$(echo $response | jq -r '.pins[0].status')
             echo "Current pin status: $pin_status"
 
--- a/.github/workflows/test-prod-e2e.yml
+++ b/.github/workflows/test-prod-e2e.yml
@@ -70,6 +70,7 @@ jobs:
     strategy:
       matrix:
         target: ["ipfs.runfission.com", "w3s.link"]
+      fail-fast: false
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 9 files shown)


laurentsenta:
Example of output: https://github.com/ipfs/gateway-conformance/actions/runs/5656544133/attempts/1#summary-15324286314

laurentsenta:
Marking this PR as ready for review, it lays down the tooling to generate dashboards for production and dev gateways.

.github/workflows/test-prod-e2e.yml
```diff
@@ -0,0 +1,141 @@
...
+          path: "gateway-conformance"
+      - name: Download Artifacts
+        uses: actions/download-artifact@v3
+        with:
+          path: artifacts
+      - name: Aggregate results
+        working-directory: ./artifacts
+        run: |
+          set -e
+          set -o pipefail
```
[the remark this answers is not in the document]
galargh: Yes, the scripts are calld as follows `shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}`

.github/workflows/update-badge.yml
```diff
@@ -0,0 +1,59 @@
+# Note: this workflow requires the repository to give Write access to Github Workflows.
+# in Settings > Actions > General > Workflow permissions
```
galargh, line 2: We can make the permission requirement explicit - https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs. If we specify them at top level, they're applied to all jobs.

```suggestion
# Note: this workflow requires the repository to give Write access to Github Workflows.
# in Settings > Actions > General > Workflow permissions

permissions:
  contents: write
```

.github/workflows/update-badge.yml
```diff
@@ -0,0 +1,59 @@
...
+        env:
+          GITHUB_JOB_SUMMARY_URL: ${{ steps.metadata.outputs.job_summary_url }}
+          REPOSITORY: ${{ github.repository }}
+      - run: |
+          if [[ -n $(git diff --shortstat 2> /dev/null | tail -n1) ]]; then
+            echo "GIT_DIRTY=1" >> $GITHUB_ENV
+          else
+            echo "GIT_DIRTY=0" >> $GITHUB_ENV
+          fi
+      - if: env.GIT_DIRTY == '1'
```
galargh, line 55: I tend to favour output over env lately as it's harder to use it by mistake. Here, it doesn't make much difference to be honest. 

```suggestion
      - id: git
         run: |
          if [[ -n $(git diff --shortstat 2> /dev/null | tail -n1) ]]; then
            echo "dirty=1" >> $GITHUB_OUTPUT
          else
            echo "dirty=0" >> $GITHUB_OUTPUT
          fi
      - if: steps.git.outputs.dirty == '1'
```

tooling/car/merge.go
```diff
@@ -26,20 +44,84 @@ func Merge(inputPaths []string, outputPath string) error {
...
+			uniqRoots[root.String()] = root
+		}
+	}
+
+	roots := make([]cid.Cid, 0)
+	for _, root := range uniqRoots {
+		roots = append(roots, root)
 	}
 
+	// Then aggregate all roots under a single one
```
galargh, line 57: 👏

tooling/car/merge.go
```diff
@@ -26,20 +44,84 @@ func Merge(inputPaths []string, outputPath string) error {
...
+	rootCid := lnk.(cidlink.Link).Cid
+
 	// Now prepare our new CAR file
-	fmt.Printf("Opening the %s file, with roots: %v\n", outputPath, roots)
-	rout, err := blockstore.OpenReadWrite(outputPath, roots)
+	fmt.Printf("Opening the %s file, with root: %v\n", outputPath, rootCid)
+	options := []carv2.Option{blockstore.WriteAsCarV1(true)}
+	rout, err := blockstore.OpenReadWrite(outputPath, []cid.Cid{rootCid}, options...)
 	if err != nil {
+		fmt.Println("Error here:a", err)
```
galargh, line 96: Is it debug log?

galargh (approved):

.github/workflows/test-prod-e2e.yml
```diff
@@ -0,0 +1,141 @@
...
+          W3STORAGE_TOKEN: ${{ secrets.W3STORAGE_TOKEN }}
+        run: |
+          set -e
+          set -o pipefail
+
+          response=$(curl --fail --oauth2-bearer "$W3STORAGE_TOKEN" --data-binary @fixtures.car "https://api.web3.storage/car")
+          cid=$(echo $response | jq -r .cid)
+          echo "Pinned fixtures to w3.storage with CID: $cid"
+          echo "cid=$cid" >> "$GITHUB_OUTPUT"
+      - name: Wait for pinning
```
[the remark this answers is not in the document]
laurentsenta: We could replace that whole action with a sleep at the moment.  None of the pinnings succeeded, and there's no way to tell what is going wrong from the API, even though the file is available online.

That might be due to some of our fixtures missing blocks on purpose: if the web3.storage API wait for the whole dag to be pinned, then it'll never succeed.

Related discussion on Slack:
https://filecoinproject.slack.com/archives/C027XP5BTGB/p1688577842308259?thread_ts=1688549599.246979&cid=C027XP5BTGB

aggregate-into-table.js
```diff
@@ -0,0 +1,86 @@
...
+
+    if (cell['fail'] > 0) {
+        return `:red_circle: (${cell['pass']} / ${cell['total']})`;
+    }
+    if (cell['skip'] > 0) {
+        return `:yellow_circle: (skipped)`;
+    }
+    if (cell['pass'] > 0) {
+        return `:green_circle: (${cell['pass']} / ${cell['total']})`;
+    }
```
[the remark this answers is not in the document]
laurentsenta: In js, it's common to drop the unnecessary else after a return, for example: https://github.com/airbnb/javascript#blocks--no-else-return

tooling/car/merge.go
laurentsenta: nice catch, thanks!

.github/workflows/update-badge.yml
laurentsenta: That make sense, thanks for the recommendation!

FWIW I'm looking into generalizing the update-badge, I'll update it soon starting from your implem in libp2p/test-plans, wip version: https://github.com/singulargarden/gateway-conformance/blob/59ace38c4519d16a5cd11c79ecbf85dd09419374/.github/workflows/update-badge.yml#L35

galargh (approved):

PR #110, opened 2023-07-14
chore: make clean and docker cleanup

lidel:
Some changes I made locally to make local dev easier:

- adds 'make clean' that removes artifacts
- improves ux when working with docker, especially when image is missing
- 'make' only builds the binary, kubo logic needs to be explicitly called


Makefile
```diff
@@ -15,8 +22,8 @@ provision-cargateway: ./fixtures.car
 	car -c ./fixtures.car &
 
 provision-kubo:
-	find ./fixtures -name '*.car' -exec ipfs dag import {} \;
-	find ./fixtures -name '*.ipns-record' -exec sh -c 'ipfs routing put --allow-offline /ipns/$$(basename -s .ipns-record "{}") "{}"' \;
+	find ./fixtures -name '*.car' -exec ipfs dag import --stats --pin-roots=false {} \;
+	find ./fixtures -name '*.ipns-record' -exec sh -c 'ipfs routing put --offline --allow-offline /ipns/$$(basename -s .ipns-record "{}") "{}"' \;
```
lidel, line 26: :thought_balloon:  iiuc this should make `put` finish instantly, and not block on putting them to 20 peers

Makefile
```diff
@@ -15,8 +22,8 @@ provision-cargateway: ./fixtures.car
 	car -c ./fixtures.car &
 
 provision-kubo:
-	find ./fixtures -name '*.car' -exec ipfs dag import {} \;
-	find ./fixtures -name '*.ipns-record' -exec sh -c 'ipfs routing put --allow-offline /ipns/$$(basename -s .ipns-record "{}") "{}"' \;
+	find ./fixtures -name '*.car' -exec ipfs dag import --stats --pin-roots=false {} \;
```
lidel, line 25: :information_source:  pinning blocks on things like missing blocks, we don't need it as GC is not enabled by default in Kubo

laurentsenta (approved):

PR #113, opened 2023-07-26
refactor: multi range request considers servers that do not return multi range

hacdias:
This PR updates the multi range request such that it counts for servers that are not able to reply with a multi range. When the client asks for a range request 3 things might happen:

1. Server replies with multipart response with `multipart/byteranges` as Content-Type. Includes both ranges.
2. Server replies with the `Content-Type` of the actual data and the `Content-Range` header with the first range. Body has first range body.
3. Server replies with `Content-Type` of actual data, no `Content-Range`. Body has full file.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests


tests/path_gateway_raw_test.go
```diff
@@ -157,3 +143,115 @@ func TestGatewayBlock(t *testing.T) {
...
+					Header("Content-Type").
+						Checks(func(v string) bool {
+							contentType = v
+							return v != ""
+						}),
+					Header("Content-Range").
+						Checks(func(v string) bool {
+							contentRange = v
+							return true
+						}),
```
hacdias, line 179: @laurentsenta these are mostly for collecting the headers. `Content-Type` needs to be present, but `Content-Range` might not be present.

tests/path_gateway_raw_test.go
```diff
@@ -157,3 +143,117 @@ func TestGatewayBlock(t *testing.T) {
...
+						Contains(string(fixture.MustGetRawData("dir", "ascii.txt")[6:17])),
+						Contains("Content-Range: bytes 0-4/31"),
+						Contains(string(fixture.MustGetRawData("dir", "ascii.txt")[0:5])),
+					)),
+			},
+		}, specs.PathGatewayRaw)
+	} else {
+		t.Error("Content-Type header did not match any of the accepted options")
+	}
+}
```
laurentsenta, line 259: One piece of advice here:
Create a test array variable, and append to it based on your if / else, then at the end call `RunWithSpecs(t, yourTests)`

Similar to what we do in [other test](https://github.com/ipfs/gateway-conformance/blob/182dd476f207776e3a95598b6b4f81d02fae7ab5/tests/redirects_file_test.go#L41),

It'll be clearer for future maintainainers what is the content of the test (the if / else) and what is part the test framework (the run...)

hacdias: Ok, I did that. Do you think I should keep the error as it is, or should that be done somewhere else?

laurentsenta: We could use a `t.FailNow()` to exit early, but I have no strong opinion here.

laurentsenta (approved):
👍

tests/path_gateway_raw_test.go
lidel, line 1: :warning:  nit: i feel testing ranges of raw block on deserialized path gateway is mixing too many concerns.
- If we test multiple range behavior on deserialized path gateway, we should use a regular unixfs file for this, as that is where they matter (video streaming, download resumes)
  - :point_right:  keep single range test here, but move multi-range tests to `path_gateway_unixfs_test.go`  and make sure we also test a single range of unixfs file there. You can get deterministic content type by using `.txt` file as a fixture
- If we test raw block behavior, then it should live in   `trustless-block-gateway` (version with `/ipfs/cid`, no path)
   - implementations of HTTP transport may not implement deserialized paths and not run `path-raw-gateway` at all, so we end up never testing behavior of range requests on trustless gateways
  - :point_right:  add range tests to `trustless_gateway_raw_test.go` 


@hacdias once  :point_right:  are  done, lgtm

lidel (approved):
@hacdias once below is addressed, lgtm

tests/path_gateway_raw_test.go
hacdias: I have:
1. Kept the single test range in `tests/path_gateway_raw_test.go`
2. Added single and multirange tests to `tests/path_gateway_unixfs_test.go` with `txt` file
3. Added single and multio range tests to `tests/trustless_gateway_raw_test.go` with `/ipfs/{cid}` path.

I am merging.

hacdias:
@laurentsenta can we get this released to unblock work on https://github.com/ipfs/boxo/pull/369? We want to run the new tests in Boxo and Kubo.

PR #114, opened 2023-07-26
feat: add ChecksAll

laurentsenta:


tests/path_gateway_raw_test.go
```diff
@@ -173,8 +173,10 @@ func TestGatewayBlockMultiRangeRequest(t *testing.T) {
 							return v != ""
 						}),
 					Header("Content-Range").
-						Checks(func(v string) bool {
-							contentRange = v
+						ChecksAll(func(v []string) bool {
+							if len(v) == 1 {
+								contentRange = v[0]
+							}
```
laurentsenta, line 179: You might want to check for cases with more than one value for the header.

Issue #123, opened 2023-08-02
Generate Implementation Dashboard


laurentsenta:
Continues the work started in https://github.com/ipfs/gateway-conformance/pull/116

- [x] reuse bifrost and kubo conformance workflows
- [x] generate a dashboard
- [x] make the dashboard human-readable
     - [x] generate nicer names
     - [x] link to related specs
- [ ] update both badges correctly
- [ ] run the workflow nightly
- [ ] The user should be able to check the execution logs when there is an error
- [ ] We want to show the version that was used to generate the results for a given implementation
- [ ] Our main output will be a dashboard in gh-pages, under a sub-path like project.io/dashboard
- [ ] We’ll solve the “naming” problem so that it’s easy to maintain the test groups names

## Phase 2:

- [ ] We might want to migrate to reusing the artifact produced by the project (”download their json”), instead of running their workflow in our CI

BigLep:
Awesome project breakdown @laurentsenta !

PR #125, opened 2023-08-02
feat: add version flag

laurentsenta:
:bomb: depends on [build-args](https://github.com/pl-strflt/docker-container-action/pull/9) in docker-container-action. Do not merge this is released and replaced in the current PR.

We want to output the version as a piece of metadata during testing. We will use this metadata during dashboard generation to signal gateways using different versions of the conformance test suite.

Example output: note the **version** in the dashboard
https://github.com/ipfs/gateway-conformance/actions/runs/5761510674/attempts/1#summary-15619660151

Supports https://github.com/ipfs/gateway-conformance/issues/123

- [x] Introduce a version flag in the CLI
- [x] Update the workflows to release with the correct version flag
- [x] Output the version as a piece of metadata during test execution, if possible, [related discussion in go project](https://github.com/golang/go/issues/43936)
- [x] Output version during dashboard generation 

> [!NOTE]
> This PR introduces metadata output, which we'll reuse for other pieces of data. This allows us to add more details to our tests, like a gateway checker id, specs id, etc.
> For now, we're using a special log format `--- META: ...`. This is a **temporary** workaround for the lack of metadata, which is being worked on in [golang/go](https://github.com/golang/go/issues/43936). We'll replace these with the official API when it's released.

### DIFF
--- a/.github/actions/extract-fixtures/action.yml
+++ b/.github/actions/extract-fixtures/action.yml
@@ -25,4 +25,5 @@ runs:
         ref: ${{ steps.github.outputs.action_sha || steps.github.outputs.action_ref }}
         dockerfile: Dockerfile
         args: extract-fixtures --directory="$OUTPUT" --merged="$MERGED"
-        build-args: --build-arg VERSION=${{ steps.github.outputs.action_ref }}
+        build-args: |
+          VERSION:${{ steps.github.outputs.action_ref }}
--- a/.github/actions/test/action.yml
+++ b/.github/actions/test/action.yml
@@ -46,7 +46,8 @@ runs:
         dockerfile: Dockerfile
         opts: --network=host
         args: test --url="$URL" --json="$JSON" --specs="$SPECS" --subdomain-url="$SUBDOMAIN" -- ${{ inputs.args }}
-        build-args: --build-arg VERSION=${{ steps.github.outputs.action_ref }}
+        build-args: |
+          VERSION:${{ steps.github.outputs.action_ref }}
     - name: Create the XML
       if: (inputs.xml || inputs.html || inputs.markdown) && (failure() || success())
       uses: pl-strflt/gotest-json-to-junit-xml@v1


Dockerfile
```diff
@@ -6,6 +6,7 @@ COPY ./go.mod ./go.sum ./
 RUN go mod download
 
 COPY . .
-RUN go build -o ./gateway-conformance ./cmd/gateway-conformance
+ARG VERSION=dev
+RUN go build -ldflags="-X github.com/ipfs/gateway-conformance/tooling.Version=${VERSION}" -o ./gateway-conformance ./cmd/gateway-conformance
```
galargh, line 10: What do you think about sourcing the metadata from git context? Similarly to what you already do in the Makefile. To me it feels more reliable - less moving parts.

```suggestion
RUN export TAG="$(git tag --points-at HEAD | sort -V | tail -n-1)" && \
  export COMMIT="$(git rev-parse HEAD --short)" && \
  export DIRTY_SUFFIX="$(test -n "`git status --porcelain`" && echo "-dirty" || true)" && \
  export VERSION="${TAG:-$COMMIT$DIRTY_SUFFIX}" && \
  go build -ldflags="-X github.com/ipfs/gateway-conformance/tooling.Version=${VERSION}" -o ./gateway-conformance ./cmd/gateway-conformance
```

For that to work, we'd only have to make sure the tags are fetched when we're building the docker image.

Also, one thing I noticed, we create tags slightly incorrectly. If you inspect `v0`, `v0.3` and `v0.3.0` now, you'll notice that `v0` and `v0.3` are on a different commit than `v0.3.0`. We should look into why that is - might need a fix in changelog based release action, I'll try to get to it today.

galargh: https://github.com/pl-strflt/changelog-driven-release/pull/9 <- I think this might be the fix for the tag mismatch. I'll test it later today.

laurentsenta: I think that'd create coupling / break single responsibility principle: instead of having a single argument, the docker build would depend on many side-effects to succeed (git binary, `.git` folder is around, repo is fetch'd as expected, etc).

PR #116, opened 2023-07-27
feat: add dev gateways with dashboards

laurentsenta:
- [x] reuse bifrost and kubo conformance workflows
- [x] generate a dashboard
- [x] make the dashboard human-readable
     - [x] generate nicer names
     - [x] link to related specs
- [ ] update both badges correctly

## Todo

At the moment, we have to hard-code the repository in the reusable workflow, this is too invasive, see the required changes in gateways:

- https://github.com/ipfs/bifrost-gateway/compare/main...singulargarden:bifrost-gateway2:main
- https://github.com/ipfs/kubo/compare/master...singulargarden:kubo:master

PR #130, opened 2023-08-04
feat: add IPIP and TestGroup Metadatas

laurentsenta:
Contributes to https://github.com/ipfs/gateway-conformance/issues/123
Related comment: https://github.com/ipfs/gateway-conformance/pull/116/files#r1277390949

To implement Dashboards, we introduce:

- The ability to define "human readable"  test groups in code, which do not depend on the test structure, these could be used to generate user-friendly tables like the one in [public-gateway-checker](https://ipfs-public-gateway-checker.on.fleek.co/).
- The ability to associate IPIPs reference with a test case and also link to subsections.

**Example Output: https://github.com/ipfs/gateway-conformance/actions/runs/5880695313/attempts/1#summary-15947714091**
Note how the tests are grouped, and how `TrustlessCarEntityBytes` links to the corresponding ipip

## Todos

- [x] Introduce the metadata APIs required
- [x] Draft a dashboard output that relies on these new specs
- [x] Implement a generator that outputs the groups, versions, and IPIPs

## Follow-ups

- [ ] demonstrate a more scalable generator that relies on turning outputs "queriable" data
- [ ] demonstrate a reverse-index that goes from an IPIP and list all the tests (when linked from the specs)
- [ ] When the need arise, we can add IPIP markers to a check (like a single header or payload test)

### DIFF
--- a/tests/trustless_gateway_car_test.go
+++ b/tests/trustless_gateway_car_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"testing"
 
+	"github.com/ipfs/gateway-conformance/tooling"
 	"github.com/ipfs/gateway-conformance/tooling/car"
 	. "github.com/ipfs/gateway-conformance/tooling/check"
 	"github.com/ipfs/gateway-conformance/tooling/helpers"
@@ -11,6 +12,8 @@ import (
 )
 
 func TestTrustlessCarPathing(t *testing.T) {
+	tooling.LogTestGroup(t, GroupTrustlessGateway)
+
 	subdirTwoSingleBlockFilesFixture := car.MustOpenUnixfsCar("trustless_gateway_car/subdir-with-two-single-block-files.car")
 	singleLayerHamtMultiBlockFilesFixture := car.MustOpenUnixfsCar("trustless_gateway_car/single-layer-hamt-with-multi-block-files.car")
 	dirWithDagCborWithLinksFixture := car.MustOpenUnixfsCar("trustless_gateway_car/dir-with-dag-cbor-with-links.car")
@@ -39,6 +42,7 @@ func TestTrustlessCarPathing(t *testing.T) {
 						Exactly().
 						InThatOrder(),
 				),
+			IPIP: IPIP402,
 		},
 		{
 			Name: "GET default CAR response of UnixFS file on a path with HAMT-sharded directory",
@@ -94,6 +98,8 @@ func TestTrustlessCarPathing(t *testing.T) {
 }
 
 func TestTrustlessCarDagScopeBlock(t *testing.T) {
+	tooling.LogTestGroup(t, GroupTrustlessGateway)
+
 	subdirTwoSingleBlockFilesFixture := car.MustOpenUnixfsCar("trustless_gateway_car/subdir-with-two-single-block-files.car")
 	singleLayerHamtMultiBlockFilesFixture := car.MustOpenUnixfsCar("trustless_gateway_car/single-layer-hamt-with-multi-block-files.car")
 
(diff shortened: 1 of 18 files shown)


laurentsenta:
Prelim output: https://github.com/ipfs/gateway-conformance/actions/runs/5761702734/attempts/1#summary-15620197820

Note the Trustless Gateway name (human readable) with a link to the corresponding IPIP

tests/trustless_gateway_car_test.go
```diff
@@ -11,6 +12,8 @@ import (
 )
 
 func TestTrustlessCarPathing(t *testing.T) {
+	tooling.LogTestGroup(t, GroupTrustlessGateway)
+
```
laurentsenta, line 16: We define test groups like this

tests/trustless_gateway_car_test.go
```diff
@@ -39,6 +42,7 @@ func TestTrustlessCarPathing(t *testing.T) {
 						Exactly().
 						InThatOrder(),
 				),
+			IPIP: IPIP402,
```
laurentsenta, line 45: We can associate an IPIP to a test, later maybe we can associate to the check itself.

tests/trustless_gateway_car_test.go
```diff
@@ -374,6 +384,9 @@ func TestTrustlessCarDagScopeAll(t *testing.T) {
 }
 
 func TestTrustlessCarEntityBytes(t *testing.T) {
+	tooling.LogTestGroup(t, GroupTrustlessGateway)
+	tooling.LogIPIP(t, IPIP402, "4.2.2")
```
laurentsenta, line 388: We can associate an ipip + it's section to a group of tests.

laurentsenta:
@lidel, @hacdias, @galargh I'm preparing the conformance test suite to generate more advanced dashboards, like the one we find in [pinning service compliance](https://ipfs-shipyard.github.io/pinning-service-compliance/) and gateway checkers.

The goal is to get your ack on how we'll associate metadata to tests like test groups and IPIPs references.

The output looks like this:
https://github.com/ipfs/gateway-conformance/actions/runs/5880695313/attempts/1#summary-15947714091
but not that this will change A LOT soon when I can generate more advance dashboards.

tests/trustless_gateway_car_test.go
hacdias: IPIPs aren't, in general, that long. I'm not sure if including their section is useful, as they all have more or less the same structure (summary, motivation, design, rationale, etc).

galargh (approved):

laurentsenta:
@lidel, thanks for the thorough review and spending the time to clarify!

I'll propose a new API to include your feedbacks regarding specs and ipips.

Regarding the high-level groups: my intent is to create 2 different concepts, "human-readable groups" metadata, and "specs" metadata.

- test groups will be used in other projects like public-gateway-checker, and they would be much more flexible. See my comment in https://github.com/ipfs/public-gateway-checker/issues/442#issuecomment-1687950302. We might merge, split, and rename these frequently.
- specs-related metadata will evolve along https://specs.ipfs.tech/ and will be MUCH more constrained. The end goal is to be able to link conformance tests from the specs.

BigLep:
Thanks for sharing.  I quickly scanned and saw some of the comments.  

From looking at https://github.com/ipfs/gateway-conformance/actions/runs/5880695313/attempts/1#summary-15947714091, things standing out:
1. Groups like "Path Gateway" should ideally link to the "Path Gateway" spec.  (I think Lidel had same comment.)
2. (Nice to have) Summarize results at the group level (e.g., "Path Gateway" is Green with X/X passing)
3. (Nice to have) Collapse the individual test results so one just sees the group and then can expand if interested?  Maybe possible with https://stackoverflow.com/questions/28607382/is-it-possible-to-create-a-toggle-switch-in-markdown ?  (Maybe this means having a table per test group?)
4. If can't have collapsed results, maybe consider different heading alignment?  For example, "Path Gateway" could be center aligned rather than right aligned?
5. Is there a reason we have a "conformance-" prefix (e.g., "conformance-ipfs.runfission.com")?
6. (Idea) While prototyping, maybe also add in ipfs.io or dweb.link to show what it looks like when something is mostly green?

This is a drive by because I'm interested in the work.  I likely won't be able to engage quickly here so don't block on me.

laurentsenta:
Thanks for all the comments, these informed the new PRs!

@BigLep I took into account your comments, it'll be great to revisit these once we're done with the first iteration of the dashboard ([wip](https://singulargarden.github.io/gateway-conformance/current/))

Closing this, replaced by:
- https://github.com/ipfs/gateway-conformance/pull/159
- https://github.com/ipfs/gateway-conformance/pull/156

PR #131, opened 2023-08-04
feat: 404 test for trustless gateway CARs

### LINKED ISSUE #126: TestTrustlessCarPathing missing test for 404 scenarios
We are missing a test for `/ipfs/cid/file` where `/cid` is valid but does not have `/file`.
In such case CAR response can be HTTP 200 and MUST include the last valid logical parent entity (`/cid`) to allow trustless client to verify the requested path does not include `/file` without  trusting the gateway.

hacdias:
Closes #126.

- Adds test for non-existing paths as defined in #126.
- This test will fail against both Kubo master and Kubo latest until https://github.com/ipfs/kubo/pull/10024 is merged.

### DIFF
--- a/tests/trustless_gateway_car_test.go
+++ b/tests/trustless_gateway_car_test.go
@@ -88,6 +88,30 @@ func TestTrustlessCarPathing(t *testing.T) {
 						InThatOrder(),
 				),
 		},
+		{
+			Name: "GET default CAR response for non-existing file",
+			Hint: `
+				CAR stream of a non-existing path must return 200 OK and all the blocks necessary
+				to traverse the path up to and including the parent of the first non-existing
+				segment of the path, in order to allow the client to verify that the request
+				path does not exist.
+			`,
+			Request: Request().
+				Path("/ipfs/{{cid}}/subdir/i-do-not-exist", subdirTwoSingleBlockFilesFixture.MustGetCidWithCodec(0x70)).
+				Query("format", "car"),
+			Response: Expect().
+				Status(200).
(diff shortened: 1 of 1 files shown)


hacdias:
@laurentsenta some workflows for `upload-fixtures` are failing. Maybe because this branch name has a slash on it? I see slashes are problematic...

lidel (approved):

tests/trustless_gateway_car_test.go
```diff
@@ -88,6 +88,30 @@ func TestTrustlessCarPathing(t *testing.T) {
...
+				Query("format", "car"),
+			Response: Expect().
+				Status(200).
+				Body(
+					IsCar().
+						IgnoreRoots().
+						HasBlocks(flattenStrings(t,
+							subdirTwoSingleBlockFilesFixture.MustGetCid(),
+							subdirTwoSingleBlockFilesFixture.MustGetCid("subdir"),
+						)...).
```
laurentsenta, line 110:
```suggestion
						HasBlocks(
							subdirTwoSingleBlockFilesFixture.MustGetCid(),
							subdirTwoSingleBlockFilesFixture.MustGetCid("subdir"),
						).
```

You don't need flatten string, it's a bandaid for mixing arrays & atoms when you call a variadic function in go,
but we're not mixing these here

laurentsenta (approved):

lidel (approved):
We may be able to merge this after smaller https://github.com/ipfs/kubo/pull/10084  lands instead.

PR #132, opened 2023-08-04
feat: test 404 on UnixFS and DAG-JSON/CBOR paths

### LINKED ISSUE #127: 404 tests for Deserialized UnixFS and  DAG-CBOR/JSON
Missing tests for 404 behavior when `/ipfs/cid/file` is requested and `/file` does not exist.

- It should produce HTTP 404 response. 
- We should have UnixFS and DAG-CBOR version, as the pathing works differently for them

See also #126  for CAR response behavior.

hacdias:
Closes #127. Adds some 404 tests as per the issue.


lidel (approved):
Thanks!

PR #133, opened 2023-08-04
fix: refs in extract fixtures and test

galargh:
This is needed because of the upstream fixes we applied to GitHub action in docker container action. Action ref output now holds ref instead of a sha like before.


hacdias (approved):

laurentsenta:
I think we should cut a release, most client repo uses the `@v3`

galargh:
@laurentsenta this doesn't affect external users.

laurentsenta:
🤦 the workflow breaks only in the case of live builds (for branch that are not released), thanks for the clarification @galargh

PR #148, opened 2023-08-21
feat: multiple skips

### LINKED ISSUE #141: Support skipping multiple tests
> (Not sure if  I was doing something wrong or maybeit is just a part of our story that needs more polishing, filling issue to discuss)

i was trying to run tests on CI with

```yaml
args: -skip 'TestGatewayCache/GET_for_/ipfs/_with_only-if-cached_succeeds_when_in_local_datastore' -skip 'TestGatewayCache/HEAD_for_/ipfs/_with_only-if-cached_succeeds_when_in_local_datastore'
```

When run, only one of them was excluded.

I worked around it with regex (since bothe differ in HEAD/GET):

```yaml
args: -skip 'TestGatewayCache/.*_for_/ipfs/_with_only-if-cached_succeeds_when_in_local_datastore'
```

 but we should make it possible for people running test suite to exclude more than one pattern, as they may want to exclude multiple unrelated ones.

laurentsenta:
Closes #141

### DIFF
--- a/.github/workflows/test-kubo-skipped-e2e.yml
+++ b/.github/workflows/test-kubo-skipped-e2e.yml
@@ -1,0 +1,81 @@
+name: Test Kubo Skipped (e2e)
+
+on:
+  workflow_dispatch:
+  push:
+    branches: 
+      - main
+  pull_request:
+
+jobs:
+  test:
+    runs-on: 'ubuntu-latest'
+    strategy:
+      fail-fast: false
+      matrix:
+        target: ['latest', 'master']
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.20.4
+      - uses: actions/checkout@v3
+        with:
+          path: 'gateway-conformance'
+      - name: Extract fixtures
+        uses: ./gateway-conformance/.github/actions/extract-fixtures
+        with:
+          output: fixtures
+      - uses: protocol/cache-go-action@v1
+      - run: go install github.com/ipfs/kubo/cmd/ipfs@${{ matrix.target }}
+        shell: bash
+        env:
+          GOPROXY: direct
+      - name: Configure Kubo Gateway
+        run: |
+          ipfs init;
+          source ./gateway-conformance/kubo-config.example.sh "$(pwd)/fixtures"
+          echo "IPFS_NS_MAP=${IPFS_NS_MAP}" >> $GITHUB_ENV
+        # note: the IPFS_NS_MAP set above will be passed the daemon
+      - uses: ipfs/start-ipfs-daemon-action@v1
+        with:
+          args: '--offline'
+          wait-for-addrs: false
+      - name: Provision Kubo Gateway
+        run: |
+          # Import car files
+          cars=$(find ./fixtures -name '*.car')
+          for car in $cars
+          do
+              ipfs dag import --pin-roots=false --stats "$car"
+          done
+
+          # Import ipns records
+          records=$(find ./fixtures -name '*.ipns-record')
+          for record in $records
+          do
+              key=$(basename -s .ipns-record "$record" | cut -d'_' -f1)
+              ipfs routing put --allow-offline "/ipns/$key" "$record"
+          done
+      - name: Run the tests
+        uses: ./gateway-conformance/.github/actions/test
+        with:
+          gateway-url: http://127.0.0.1:8080
+          subdomain-url: http://example.com
+          json: output.json
+          xml: output.xml
+          html: output.html
+          markdown: output.md
+          skips: '["TestTrustlessCarPathing", "TestNativeDag/GET_plain_JSON_codec_from_.*"]'
+      - name: Set summary
+        if: (failure() || success())
+        run: cat ./output.md >> $GITHUB_STEP_SUMMARY
+      - name: Upload one-page HTML report
+        if: (failure() || success())
+        uses: actions/upload-artifact@v3
+        with:
+          name: conformance-${{ matrix.target }}.html
+          path: ./output.html
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ The `test` command is the main command of the tool. It is used to test a given I
 | html | GitHub Action | The path where the one-page HTML test report should be generated. | `./report.html` |
 | markdown | GitHub Action | The path where the summary Markdown test report should be generated. | `./report.md` |
 | specs | Both | A comma-separated list of specs to be tested. Accepts a spec (test only this spec), a +spec (test also this immature spec), or a -spec (do not test this mature spec). | Mature specs only |
+| skip | Both | Run only the those tests that do not match the regular expression. Similar to golang's skip, the expression is split by slash (/) into a sequence and each part must match the corresponding part in the test name, if any | empty |
 | args | Both | [DANGER] The `args` input allows you to pass custom, free-text arguments directly to the Go test command that the tool employs to execute tests. | N/A |
 
 ##### Specs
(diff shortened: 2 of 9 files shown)


.github/workflows/test-kubo-skipped-e2e.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Run the tests
+        uses: ./gateway-conformance/.github/actions/test
+        with:
+          gateway-url: http://127.0.0.1:8080
+          subdomain-url: http://example.com
+          json: output.json
+          xml: output.xml
+          html: output.html
+          markdown: output.md
+          skips: '["TestTrustlessCarPathing", "TestNativeDag/GET_plain_JSON_codec_from_.*"]'
```
laurentsenta, line 72: Highlight: this is how we skip tests. Note the regexp similar to golang's matching

README.md
```diff
@@ -235,10 +237,11 @@ gateway-conformance test --specs trustless-gateway,-trustless-gateway-ipns
 
 ### Skip a specific test
 
-Tests are skipped using Go's standard syntax:
+Tests are skipped using the `--skip` parameter and Go's standard syntax:
 
 ```bash
-gateway-conformance test -skip 'TestGatewayCar/GET_response_for_application/vnd.ipld.car/Header_Content-Length'
+gateway-conformance test --skip 'TestGatewayCar/GET_response_for_application/vnd.ipld.car/Header_Content-Length'
+gateway-conformance test --skip 'TestGatewayCar/.*/vnd.ipld.car' --skip 'TestGatewayCar/GET_response_for_application/vnd.*'
```
laurentsenta, line 244: highlight: use in CLI

.github/workflows/test-kubo-skipped-e2e.yml
galargh: It's the same as the following, right?

```suggestion
          args: '-skip "(TestTrustlessCarPathing|TestNativeDag/GET_plain_JSON_codec_from_.*)"'
```

README.md
```diff
@@ -235,10 +237,11 @@ gateway-conformance test --specs trustless-gateway,-trustless-gateway-ipns
 
 ### Skip a specific test
```
lidel, line 238: Could  we add `### Run only a specific test` with example that uses go's `go test -run <pattern>` and use the same pattern as in the skip example?

.github/workflows/test-kubo-skipped-e2e.yml
lidel: Regex support is pretty recent, i think? https://tip.golang.org/doc/go1.20 states:

> The go test command now accepts -skip <pattern> to skip tests, subtests, or examples matching <pattern>.

`go help testflag` conforms  we now have parity for `-run` and -skip`:

```
-run regexp
	    Run only those tests, examples, and fuzz tests matching the regular
	    expression. For tests, the regular expression is split by unbracketed
	    slash (/) characters into a sequence of regular expressions, and each
	    part of a test's identifier must match the corresponding element in
	    the sequence, if any. Note that possible parents of matches are
	    run too, so that -run=X/Y matches and runs and reports the result
	    of all tests matching X, even those without sub-tests matching Y,
	    because it must run them to look for those sub-tests.
	    See also -skip.
[..]
	-skip regexp
	    Run only those tests, examples, fuzz tests, and benchmarks that
	    do not match the regular expression. Like for -run and -bench,
	    for tests and benchmarks, the regular expression is split by unbracketed
	    slash (/) characters into a sequence of regular expressions, and each
	    part of a test's identifier must match the corresponding element in
	    the sequence, if any.
```

Both `skip` and `run` are useful.  `run` allows more fine-grained control than our predefined presets, which saves time when someone develops a fix and has to re-run specific subset of tests without running the entire suite (not a problem today, but the suite will grow).

@laurentsenta  do we need to have dedicated input in github action? 
If we could keep this under `args` we would not have to do anything special to support `-run` here, right?

cmd/gateway-conformance/main.go
```diff
@@ -113,6 +115,13 @@ func main() {
 						Value:       "",
 						Destination: &specs,
 					},
+					&cli.StringSliceFlag{
+						Name: "skip",
+						Usage: "Accepts a test path to skip.\nCan be used multiple times.\n" +
+							"Example: --skip \"TestTar/GET_TAR_with_format=.*/Body\" --skip \"TestGatewayBlock\" --skip \"TestTrustlessCarPathing\"\n" +
+							"It uses the same syntax as the -skip flag of go test.",
+						Destination: &skips,
+					},
```
lidel, line 124: Parity with  `go-test -skip` is :+1:
Could we include the same for `go-test -run` in this PR? 
It will make things feature-complete, and natural for anyone familiar with go, and at least remove surprises when someone tries `-skip` and `-run` – these should work the same, and if we support passing `-skip` multiple times, we should also support that for `-run` :pray:

lidel (changes requested):
Thank you @laurentsenta!

Details inline, quick feedback:
- add support for passing multiple `-run` (for parity with `-skip`)
- could we keep `args` in github action examples? (this way we don't need special casing for `runs`)

README.md
```diff
@@ -235,10 +237,11 @@ gateway-conformance test --specs trustless-gateway,-trustless-gateway-ipns
...
 
-Tests are skipped using Go's standard syntax:
+Tests are skipped using the `--skip` parameter and Go's standard syntax:
```
lidel, line 240:
```suggestion
Tests can be skipped using the `--skip` parameter which follows `go test -skip regex` syntax, but can be passed more than once for convenience:
```

.github/workflows/test-kubo-skipped-e2e.yml
galargh: I'm also wondering if we even need dedicated run/skip gateway conformance cli flags if we can accomplish the same using the go flags.

If we do, e.g. as nicer to use sugar, I think it might be worth considering converting the custom flag values into regex and passing it to go to avoid having to reimplement run/skip logic.

laurentsenta: 👍 Thanks for the reviews and comments,
That work might be a dead end, Sorry for the noise 🤦 

I ran more experiments, and not everything makes sense to me, but regular skips should be enough.

@galargh your suggestion would need a different parenthesizing:

```
(TestTrustlessCarPathing|TestNativeDag/GET_plain_JSON_codec_from_.*) # does not work
(TestTrustlessCarPathing)|(TestNativeDag/GET_plain_JSON_codec_from_.*) # would work
```

@lidel your issue should be solved with an `|` and parenthesis:
```sh
go test ... -skip '(TestGatewayCache/GET_for_/ipfs/_with_only-if-cached_succeeds_when_in_local_datastore)|(TestGatewayCache/HEAD_for_/ipfs/_with_only-if-cached_succeeds_when_in_local_datastore)'
```

Probably also:
```sh
go test ... -skip 'TestGatewayCache/(GET|HEAD)_for_/ipfs/_with_only-if-cached_succeeds_when_in_local_datastore'
```

We can drop that PR.

I think the easiest way to think about golang `-skip` is:
- create groups with `(g1)|(g2)` at the top level,
- then each group is split by `/` and each part is matched with the go test name.
   - `willSkip("regexp1/regexp2", "testX/testY") = match(regexp1, testX) && match(regexp2, testY)`

I'm still confused by some aspects of the matching... if you want to dig deeper: https://gist.github.com/laurentsenta/5ed9fe9954605eefc1ed9ee2db7f66a9

But I'll close this PR for now and continue the discussion in #141.

galargh: Thank you for looking into it 🙇 That's a nice write-up for a tricky feature 👏

galargh:
@laurentsenta Should we close it?

galargh:
Closing it now.

PR #152, opened 2023-08-25
feat: generate the Dashboard - base structure

laurentsenta:
example output:
https://singulargarden.github.io/gateway-conformance/

example workflow:
https://github.com/singulargarden/gateway-conformance/actions/workflows/deploy-pages.yml

# Adds data generation

- `munge.js`: takes a go test output and generates a nicer data structure
- `munge_sql.js`: load all the outputs into a single sqlite db
- `munge_aggregates.js`: generate resources for the website generation (content pages and data files)

## Generate the dashboard

- in www/
- using Hugo

## Follow-ups (see https://github.com/ipfs/gateway-conformance/issues/123)

- [ ] complete drill down (fix slugs, pathing, and breadcrumbs)
- [ ] complete logs rendering (nesting for logs)
- [ ] more prominent jobs urls


munge_sql.js
```diff
@@ -0,0 +1,168 @@
+/**
+ * This script takes the munged JSON files and puts them into a SQLite database.
+ */
+const sqlite3 = require('sqlite3').verbose();
```
galargh, line 4: What does verbose do? Do we want it always on or only when debugging? FYI, `RUNNER_DEBUG=1` env var indicates that a workflow run is being rerun with debugging logging enabled (if that's useful).

munge_sql.js
```diff
@@ -0,0 +1,168 @@
...
+}
+
+main()
+    .then(() => {
+        console.log("done");
+    })
+    .catch((e) => {
+        console.error(e);
+        process.exit(1);
+    })
```
galargh, line 168:
```suggestion
    })

```

package.json
```diff
@@ -0,0 +1,19 @@
...
+  "dependencies": {
+    "gray-matter": "^4.0.3",
+    "sqlite3": "^5.1.6"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC"
```
galargh, line 18: Just flagging that it was set by default.

www/themes/conformance/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "npx tailwindcss -i ./tailwind.css --postcss ./postcss.config.js -o ./assets/style.css",
+    "watch": "npx tailwindcss -i ./tailwind.css --postcss ./postcss.config.js -o ./assets/style.css --watch"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
galargh, line 12: Same here.

galargh (approved):
FYI, I didn't review SQL too closely - assuming it's OK if it works. And HTML, I'd say that the visual review we did together during demo last time is good enough. All in all, good to go from my side :)

.github/workflows/deploy-pages.yml
```diff
@@ -86,29 +86,33 @@ jobs:
         with:
           path: artifacts
       - name: Generate Data Aggregates
-        working-directory: ./artifacts
+        working-directory: ./
         run: |
-          mkdir ../aggregates
+          npm ci
+          mkdir ./munged
```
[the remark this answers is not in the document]
laurentsenta: It's used in [data science](https://www.talend.com/resources/what-is-data-munging/)
> Data munging is a set of concepts and a methodology for taking data from unusable and erroneous forms to the new levels of structure and quality required by modern analytics processes and consumers. 

We can use a different name though, no strong feeling about this.

munge.js
```diff
@@ -0,0 +1,129 @@
+/**
+ * This file loads a test2json output from stdin,
+ * and generate a test result structure.
+ * 
+ * The output is a map of test names to test results.
+ * 
+ * A test result is an object with the following fields:
```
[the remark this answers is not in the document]
laurentsenta: Yup, I'd rather keep it simple for now, but eventually... I also considered turning this into a go binary and using test2json libs.

munge.js
```diff
@@ -0,0 +1,129 @@
...
+    .filter((line) => line !== "")
+    // ## extract json
+    .map((line) => {
+        try {
+            return JSON.parse(line);
+        } catch (e) {
+            throw new Error(`Failed to parse line: ${line}: ${e}`);
+        }
+    })
+    // ## Drop lines that are not about a Test
```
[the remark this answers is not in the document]
laurentsenta: Packages:
```json
{"Time":"2023-08-31T09:45:33.640286+02:00","Action":"start","Package":"Gateway Tests"}
```

munge_aggregates.js
```diff
@@ -0,0 +1,325 @@
...
+    const content = {
+        content: "",
+        data: {}
+    }
+    if (fs.existsSync(fullPath)) {
+        const existing = matter.read(fullPath);
+        content.content = existing.content;
+        content.data = existing.data;
+    }
+    content.data = { ...content.data, ...data };
```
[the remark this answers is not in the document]
laurentsenta: So we'll have to garbage collect at some point, but I'd like to preserve the files instead of overwriting:
this way, we can have humans add content and details, since each page (like "kubo", or "test-XYZ") is a regular page.

munge_sql.js
laurentsenta: We'll get more detailed stack traces on exceptions, which we'll be helpful even without debug logging enabled.
https://github.com/TryGhost/node-sqlite3/wiki/API#sqlite3verbose

package.json
laurentsenta: Ho yea, we need to find a license at some point. Dropped for now.

BigLep:
Very nice!

A few things coming to mind:
1. Do we have docs describing the flow of the dashboard generation?  For example, I can't remember, are we pulling the conformance results from the CIs of other projects (like Kubo, Boxo) or are we running the conformance tests ourselves in this repo?  (it looks like we pull the results per https://github.com/singulargarden/gateway-conformance/actions/runs/6075886236 ).  It would be good to have docs on "how to add your implementation to the dashboard".  I think that will cause the right high level info to surface.
2. Assuming we're pulling results from other repos, what's the way for someone to see:
   - what commit of the conformance test suite we ran
   - what commit of the implementation we ran against
   - date when these results are from
Is that all in the job log?

Good stuff.  

I would also talk with @lidel on how we give visibility to this dashboard from https://specs.ipfs.tech/

laurentsenta:
@BigLep thanks for the nice and detailed feedback, I'll update https://github.com/ipfs/gateway-conformance/issues/123 to take into account your notes.

PR #157, opened 2023-09-07
feat: ipns v2 and v2 record combination and tests

### LINKED ISSUE #144: Add more IPNS record tests based on IPIP-428
Ref. https://github.com/ipfs/specs/pull/428 tldr: we should have `.ipns-record` files with fixtures for:

- [ ]  V1-only → record invalid (gateway returns 5XX error)
- [ ] V1+V2 (both signatures valid) → record valid, HTTP 200
- [ ]  V1+V2 (both signatures valid, but 'value' is different in V1 pb) → record invalid (gateway returns 5XX error)
- [ ] V1+V2 (only signatureV1 valid) → record invalid (gateway returns 5XX error)
- [ ] V1+V2 (only signatureV2 valid) → record valid, HTTP 200
- [ ] V2-only (no V1 fields) → record valid, HTTP 200 
- [ ] update [IPIP-428](https://github.com/ipfs/specs/pull/428) and include links to mentioned fixtures from this PR

This will be an extremely useful resource for anyone trying to implement IPNS, as they would  be able to either run gateway-conformance, or at least reuse related fixtures.

hacdias:
Closes #144. Many IPNS tests. This fixtures can also later be re-used for when we add TTL. It also sets a nice precedent of adding a file where we programatically generate the fixtures in a language other than shell.


tooling/test/sugar.go
```diff
@@ -152,6 +154,12 @@ func (e ExpectBuilder) Status(statusCode int) ExpectBuilder {
 	return e
 }
 
+func (e ExpectBuilder) StatusRange(from, to int) ExpectBuilder {
```
laurentsenta, line 157: Would `StatusBetween` be more self-explanatory?

fixtures/ipns_records/README.md
```diff
@@ -4,20 +4,13 @@
...
-# IPNS_KEY=k51qzi5uqu5dh71qgwangrt6r0nd4094i88nsady6qgd1dhjcyfsaqmpp143ab
-# FILE_CID=bafkreidfdrlkeq4m4xnxuyx6iae76fdm4wgl5d4xzsb77ixhyqwumhz244 # A file containing Hello IPFS
-```
+k51qzi5uqu5dm4tm0wt8srkg9h9suud4wuiwjimndrkydqm81cqtlb5ak6p7ku_v1.ipns-record -> /ipfs/bafkqadtwgeww63tmpeqhezldn5zgi
+k51qzi5uqu5dlkw8pxuw9qmqayfdeh4kfebhmreauqdc6a7c3y7d5i9fi8mk9w_v1-v2.ipns-record -> /ipfs/bafkqaddwgevxmmraojswg33smq
+k51qzi5uqu5dlmit2tuwdvnx4sbnyqgmvbxftl0eo3f33wwtb9gr7yozae9kpw_v1-v2-broken-v1-value.ipns-record -> /ipfs/bafkqahtwgevxmmraojswg33smqqho2lunaqge4tpnnsw4idwmfwhkzi
+k51qzi5uqu5dilgf7gorsh9vcqqq4myo6jd4zmqkuy9pxyxi5fua3uf7axph4y_v1-v2-broken-signature-v1.ipns-record -> /ipfs/bafkqahtwgevxmmrao5uxi2bamjzg623fnyqhg2lhnzqxi5lsmuqhmmi
+k51qzi5uqu5diamp7qnnvs1p1gzmku3eijkeijs3418j23j077zrkok63xdm8c_v1-v2-broken-signature-v2.ipns-record -> /ipfs/bafkqahtwgevxmmrao5uxi2bamjzg623fnyqhg2lhnzqxi5lsmuqhmmq
+k51qzi5uqu5dit2ku9mutlfgwyz8u730on38kd10m97m36bjt66my99hb6103f_v2.ipns-record -> /ipfs/bafkqadtwgiww63tmpeqhezldn5zgi
+```
```
laurentsenta, line 16: If we have to fix & run this code one 2 years from now for example, is there a way to tell which parts of the codebase to update?

tests/path_gateway_ipns_test.go
```diff
@@ -7,6 +7,74 @@ import (
...
+	bodyIPNSV1V2BrokenSigV1 = []byte("v1+v2 with broken signature v1")
+
+	ipnsV1V2     = "k51qzi5uqu5dlkw8pxuw9qmqayfdeh4kfebhmreauqdc6a7c3y7d5i9fi8mk9w"
+	bodyIPNSV1V2 = []byte("v1+v2 record")
+	cidIPNSV1V2  = "bafkqaddwgevxmmraojswg33smq"
+
+	ipnsV2     = "k51qzi5uqu5dit2ku9mutlfgwyz8u730on38kd10m97m36bjt66my99hb6103f"
+	bodyIPNSV2 = []byte("v2-only record")
+	cidIPNSV2  = "bafkqadtwgiww63tmpeqhezldn5zgi"
+)
```
laurentsenta, line 26: Initially, the goal was to move around car files so that we could store them elsewhere (in the specs repo for example). We'd reduce dependencies on hardcoded values as we had in sharness.

That's why we wrote fixtures loaders like:

```go
	fixture := car.MustOpenUnixfsCar("ipns_records/fixtures.car")
	file := fixture.MustGetRoot()
	fileCID := file.Cid()

	ipns := MustOpenIPNSRecordWithKey("ipns_records/k51qzi5uqu5dh71qgwangrt6r0nd4094i88nsady6qgd1dhjcyfsaqmpp143ab.ipns-record")
	ipnsName := ipns.Key()
```

or

https://github.com/ipfs/gateway-conformance/blob/fd7e8c954ed7f38e3a01ced47f23929b7108cfed/tests/subdomain_gateway_ipfs_test.go#L108-L116


- Eventually, we'd use assets (car files) coming from the specs, and the test-suite would update "declaratively" and throw an error as soon as you'd load the test (because the file would be missing from a car file for example).
- We also discussed having ipns records in car files, which meant we could remove hardcoding keys in filenames.


To me this seems to be a step back, so I'd rather make sure we're making this change consciously.

hacdias: @laurentsenta I'm making this very consciously. Most of these records are invalid, and therefore `ipns.UnmarshalRecord` will fail. So we can't load them like that.

tests/path_gateway_ipns_test.go
```diff
@@ -7,6 +7,74 @@ import (
...
+				Path("/ipns/{{name}}", ipnsV2),
+			Response: Expect().
+				Body(bodyIPNSV2),
+		},
+		{
+			Name: "GET an IPNS Path (V1+V2) with broken SignatureV2 from the gateway fails with 5xx",
+			Request: Request().
+				Path("/ipns/{{name}}", ipnsV1V2BrokenSigV2),
+			Response: Expect().
+				StatusBetween(500, 599),
```
laurentsenta, line 71: We could implement a `.Status("5xx")` if you need

tests/path_gateway_ipns_test.go
laurentsenta: These could use the regular API, right? 

```go
	ipnsV1V2     = "k51qzi5uqu5dlkw8pxuw9qmqayfdeh4kfebhmreauqdc6a7c3y7d5i9fi8mk9w"
	bodyIPNSV1V2 = []byte("v1+v2 record")
	cidIPNSV1V2  = "bafkqaddwgevxmmraojswg33smq"

	ipnsV2     = "k51qzi5uqu5dit2ku9mutlfgwyz8u730on38kd10m97m36bjt66my99hb6103f"
	bodyIPNSV2 = []byte("v2-only record")
	cidIPNSV2  = "bafkqadtwgiww63tmpeqhezldn5zgi"
```

tests/path_gateway_ipns_test.go
hacdias: @laurentsenta that is also an option

tests/path_gateway_ipns_test.go
hacdias: @laurentsenta, yes, I can update the ones that can use the regular API, I'll update them

hacdias: @laurentsenta updated.

laurentsenta (approved):
lgtm, thanks for applying the fixes.

lidel (approved):
Thank you @hacdias!  Lgtm. I will refer to these fixtures from [IPIP-428](https://github.com/ipfs/specs/pull/428).

I've added missing `Hint` to some V1/V2 test cases to make it self-explanatory for the reader why certain checks exist.

@laurentsenta I've resolved conflict with `main`, ok to merge?

laurentsenta:
@lidel merged, thanks for the review and conflict resolution! @hacdias thanks for adding the test

PR #159, opened 2023-09-07
feat: add specs and user taxonomies

laurentsenta:
Contributes to #123 

Add a "User Taxonomy" different from the specs. This will be used in gateway checker and other user-facing tooling.

- [x] Introduce API
- [x] Add "most" tests to groups

Add specs that connect with specs.ipfs.tech

- [x] add spec(s)
- [x] update (some) tests with these
- [x] use these in the outputs (at least in markdown file)

⚠️ I'm not trying to assign specs and groups to all the tests for now, just demonstrate the specs API. We'll build the rest of the dashboard, maybe even a spec coverage table first.
 
Follow-ups:

- [ ] use these in dashboard generation
- [ ] validate these at build time by checking the pages & anchor exists

### DIFF
--- a/tests/path_gateway_dag_test.go
+++ b/tests/path_gateway_dag_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 func TestGatewayJsonCbor(t *testing.T) {
-	tooling.LogTestGroup(t, GroupPathGateway)
+	tooling.LogTestGroup(t, GroupJSONCbor)
 
 	fixture := car.MustOpenUnixfsCar("path_gateway_dag/gateway-json-cbor.car")
 
@@ -69,7 +69,7 @@ func TestGatewayJsonCbor(t *testing.T) {
 // ## Reading UnixFS (data encoded with dag-pb codec) as DAG-CBOR and DAG-JSON
 // ## (returns representation defined in https://ipld.io/specs/codecs/dag-pb/spec/#logical-format)
 func TestDagPbConversion(t *testing.T) {
-	tooling.LogTestGroup(t, GroupPathGateway)
+	tooling.LogTestGroup(t, GroupJSONCbor)
 
 	fixture := car.MustOpenUnixfsCar("path_gateway_dag/gateway-json-cbor.car")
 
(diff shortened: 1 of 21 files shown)


laurentsenta:
Example output: https://github.com/ipfs/gateway-conformance/actions/runs/6119761024/attempts/1#summary-16610661993

tests/path_gateway_dag_test.go
```diff
@@ -541,6 +553,7 @@ func TestNativeDag(t *testing.T) {
 			},
 			{
 				Name: Fmt("HEAD {{name}} with only-if-cached for missing block returns HTTP 412 Precondition Failed", row.Name),
+				Spec: "specs.ipfs.tech/http-gateways/path-gateway/#only-if-cached",
```
lidel, line 556: @laurentsenta  lgtm, but cosmetic nit: dropping `https:` does not buy us much, could we  keep full URLs?
It will enable us to use automated tools that find URLs in strings and report broken ones without extra step, users will be able to click the URL in their code editor etc.

tests/path_gateway_tar_test.go
```diff
@@ -73,6 +76,7 @@ func TestTar(t *testing.T) {
 		},
 		{
 			Name: "GET TAR with explicit ?filename= succeeds with modified Content-Disposition header",
+			Spec: "specs.ipfs.tech/http-gateways/path-gateway/#content-disposition-response-header",
```
lidel, line 79: @laurentsenta thoughts on having `Specs` version which allows for more than 1 URL?  I imagine there will be cases where more than one spec is relevant.

tests/path_gateway_dag_test.go
laurentsenta: Will fix, thanks!

tests/path_gateway_tar_test.go
laurentsenta: Thought about it, you can use Specs! https://github.com/ipfs/gateway-conformance/pull/159/files#diff-73ca403802a777f0a38d9aed0daa6f62819960272f8997e91c7eb6bd6782935fR17

PR #160, opened 2023-09-08
test: percent-encoded filename and  _redirects with If-None-Match on DNSLink

### LINKED ISSUE #111: Add test for _redirects with  If-None-Match headers
We need a regression test for https://github.com/ipfs/boxo/pull/412

iiuc it should have `_redirects` 

```
# Map SPA routes to the main index HTML file.
/* /index.html 200
```

and make request for `/ipfs/cid/MISSING-PAGE` twice, second time sending `If-None-Matc` with `Etag` value from the first response.

There is a test already on the linked PR, and aCAR fixture which we should reuse as-is:
Ref. https://github.com/ipfs/boxo/pull/412/files#diff-d34c7069ce5e81d45082b19eb3e869ee1a086e185dcd6630e75e3ed0d368b546

### LINKED ISSUE #115: Add test for content paths that have percent-encoded parts
> Found this edge case in https://github.com/ipfs/bifrost-gateway/pull/160

Content Paths with special characters are usually encoded before sending to the gateway.
However there is an edge case where a content path has already encoded parts.
We do not want it to be mangled during transport OR things like subdomain redirects.

There should be a conformance test based on below example (we dont want to pull-in entire wikipedia, but its fine to reuse the single file from `/I` directory as-is).

## Example

The original content path is `/ipns/en.wikipedia-on-ipfs.org/I/Auditorio_de_Tenerife%2C_Santa_Cruz_de_Tenerife%2C_España%2C_2012-12-15%2C_DD_02.jpg.webp` (note `ñ` in `España%2C`)

But the URL path used in HTTP request is percent-encoded:

`curl -v "https://en-wikipedia--on--ipfs-org.ipns.dweb.link/I/Auditorio_de_Tenerife%252C_Santa_Cruz_de_Tenerife%252C_Espa%C3%B1a%252C_2012-12-15%252C_DD_02.jpg.webp"`  

Test with curl (incl. subdomain redirect):

`curl -v -L "http://localhost:8081/ipns/en.wikipedia-on-ipfs.org/I/Auditorio_de_Tenerife%252C_Santa_Cruz_de_Tenerife%252C_Espa%C3%B1a%252C_2012-12-15%252C_DD_02.jpg.webp"`

For now, I've added basic unit test in https://github.com/ipfs/bifrost-gateway/commit/982535a97b046017cc320ddb08a9fcb44c2a1b94#diff-84d756d48159dd2e742a98880c23ce6d22bd80bf878409ddcf1d4dc81bc114acR38, but we should have a conformance test for this case.

hacdias:
- Closes #115.
- Closes #111.


hacdias:
@laurentsenta I'm not sure what i did wrong in b88910b but the DNSLink tests are not working and just getting `no such host`.

laurentsenta:
@hacdias Use `helpers.UnwrapSubdomainTests(t, tests)`
https://github.com/ipfs/gateway-conformance/blob/bdeb1a993c02027cff92145ac07f943918f959a9/tests/redirects_file_test.go#L279

hacdias:
Thanks @laurentsenta. It works now.

@lidel  this is now complete. I won't be working this week, so feel free to take it over and merge it.

laurentsenta:
@lidel @hacdias I'm going to trigger a release today, should we merge this?

lidel (approved):
@hacdias @laurentsenta fine to merge and include in release, but if you miss it also ok (as these are not as critical for Rhea/Saturn as range requests).

laurentsenta:
Thanks for the contributions and reviews 🙏

Issue #154, opened 2023-08-29
Add tests for HTTP Range requests


lidel:
## TLDR

Coverage of range behaviors is bare minimum for UnixFS right now.
We should improve it and both relax requirements (allowing implementers to choose what to support) and make asserts more explicit (if you choose to implement something optional, you should behave A, if you don't, we expect fallback to be either B or C etc).

When it comes to range requests, HTTP specs ([RFC9110](https://httpwg.org/specs/rfc9110.html#range.requests)) make them OPTIONAL. Supporting HTTP Ranges is useful for streaming chunks of deserialized videos, and that is what some IPFS Gateways support when a client asks for deserialized data (a single range).

## Missing tests

This means these behaviors may occur:
- returned HTTP 206 with bytes for the first range
- returned HTTP 206 with content type `multipart/byteranges` and multiple ranges 
  - this is not used in practice, browsers streaming videos ask for a single range, Kubo >=0.23 does not support this for unixfs.
- returned HTTP 200 with the whole file 
- returned HTTP 200 with the whole file

a file could be any type of deserialized response, but in practice end users only care about UnixFS file

We don't want to be too strict here, but also we want to be able to write a test that limits the possibility space to the above paths.
If someone implements Ranges, we want to ensure the response is valid.
If they don't we also want to ensure the response is valid.

## What we need in DSL

It is most likely that we need to extend test DSL with something like `AnyOf` described in https://github.com/ipfs/gateway-conformance/issues/153

lidel:
We have tests for trustless `?format=car&entity-bytes=0:0` but we are lacking a similar test for deserialized responses with HTTP header (`Range: bytes=0-0`).

- [ ] **HTTP Range request works and does overfetch**
  - Reuse [gateway-conformance/fixtures/trustless_gateway_car/file-3k-and-3-blocks-missing-block.car](https://github.com/ipfs/gateway-conformance/raw/ed74b532fc7662732ec87d92c49435668369db66/fixtures/trustless_gateway_car/file-3k-and-3-blocks-missing-block.car): A file chunked into 3 blocks, each 1024 bytes. The middle block is missing to ensure range request only triggers fetch of necessary blocks.
  - HTTP `Range: bytes=0-0` request for `/ipfs/QmYhmPjhFjYFyaoiuNzYv8WGavpSRDwdHWe5B4M5du5Rtk`, expect response to be HTTP 206 with    body equal to the first byte of the file
[...]

Issue #153, opened 2023-08-29
Support MUST / MAY / SHOULD responses


laurentsenta:
A spec might allow for multiple responses depending on the gateway's capability.

For example, the [specs for path gateway - range request header](https://specs.ipfs.tech/http-gateways/path-gateway/#range-request-header) allows a gateway to return either:

- the minimal dag matched by a range request using multiple ranges,
- the minimal dag matched by the first range request,
- the whole dag.

We solve this by running a first check that requests an endpoint, stores the response's header through a side-effect, then uses an `if / else` flow to test between these capabilities ([code](https://github.com/ipfs/gateway-conformance/blob/1f6a48c4d10bee567bdbecce65e7f640793768d1/tests/path_gateway_unixfs_test.go#L456-L560)).

@aschmahmann is working on adding more of these tests across the suite; before implementing many more, we might want to support this natively through the sugar API.

Also, since two gateways might pass the specs with very different feature-sets, it would make sens to signal which features are support to the end user/gateway implementers.

## Problem 1: Support many possible responses

At the moment you may define a test with:

```golang
{
   Name: ...
   Request: Request().Path(...)....
   Response: Expect().Headers(...)....
},
{
   Name: ...
   Requests: Requests(Request()...., Request()..., ....) // multiple requests
   Response: Expect().Headers(...)....
   Responses: Responses().HaveTheSamePayload(),... // checker for multiple responses
}
```

We could add `Either` operator:

```golang
{
   Name: ...
   Request: Request().Path(...)....
   Response: Either(
       Response()...., // returned multiple ranges
       Response()...., // returned first range
       Response().... // returned whole file
   )
}
```

That would pass if the actual response matches any of the responses.
We could implement this as a /simple/ "Or" operator.

This opens up more complex operators, like

```golang
{
   Name: ...
   Request: Request().Path(...)....
   Response: And(
       Response()...., // shared response expectation,
       Either(
           Response()...., // returned multiple ranges
           Response()...., // returned first range
           Response().... // returned whole file
       )
   )
}
```

## Problem 2: Surface MUST / MAY / SHOULD features

Rough idea:
- If a spec is a MUST, the test should PASS or FAIL if the expected responses is matched
- If a spec is a MAY or SHOULD, the test should PASS or SKIP if the expected responses is matched

This will signal to users/implementers which features are supported or not.

[...]

Issue #36, opened 2023-04-10
make sure builders / expectations API do not mutate the input struct


laurentsenta:
Not all APIs are pure right now:

- for example, the `tar` API mutates the builder object.
- some requests and header checks are also doing side effects (with the header map for example).

We want to make sure these do not mutate the input object so that a pattern like this is correct:

```go
req1 := SomeCheck().A().B()
req2 := req1.C().D()
```

PR #162, opened 2023-09-11
feat: range request helpers

aschmahmann:
- Add tests for #154 
- Adds basic `AnyOf` and `AllOf` helpers (related to #153)
- Range request helpers
- Added range requests tests for IPLD block requests

cc @laurentsenta @lidel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added tests for HTTP Range requests, as well as some basic helpers for `AnyOf` and `AllOf`. [PR](https://github.com/ipfs/gateway-conformance/pull/162)
 ## [0.3.1] - 2023-09-15
 ### Added
 - Specs Dashboard Output. [PR](https://github.com/ipfs/gateway-conformance/pull/163)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
 ### Added
-
 - Added tests for HTTP Range requests, as well as some basic helpers for `AnyOf` and `AllOf`. [PR](https://github.com/ipfs/gateway-conformance/pull/162)
+
 ## [0.3.1] - 2023-09-15
 ### Added
 - Specs Dashboard Output. [PR](https://github.com/ipfs/gateway-conformance/pull/163)
--- a/tests/path_gateway_dag_test.go
+++ b/tests/path_gateway_dag_test.go
@@ -234,25 +234,25 @@ func TestPlainCodec(t *testing.T) {
 
 		var dagFormattedResponse []byte
 
-		tests := &SugarTests{}
-		tests.Append(
-			helpers.IncludeRandomRangeTests(t,
-				SugarTest{
-					Name: Fmt(`GET {{name}} without Accept or format= has expected "{{format}}" Content-Type and body as-is`, row.Name, row.Format),
-					Hint: `
+		tests := SugarTests{}.
+			Append(
+				helpers.IncludeRandomRangeTests(t,
+					SugarTest{
+						Name: Fmt(`GET {{name}} without Accept or format= has expected "{{format}}" Content-Type and body as-is`, row.Name, row.Format),
+						Hint: `
 				No explicit format, just codec in CID
 				`,
-					Request: Request().
-						Path("/ipfs/{{cid}}", plainCID),
-					Response: Expect().
-						Headers(
-							Header("Content-Disposition").
-								Contains(Fmt(`{{disposition}}; filename="{{cid}}.{{format}}"`, row.Disposition, plainCID, row.Format)),
-						),
-				},
-				plain.RawData(),
-				Fmt("application/{{format}}", row.Format),
-			)...).
+						Request: Request().
+							Path("/ipfs/{{cid}}", plainCID),
+						Response: Expect().
+							Headers(
+								Header("Content-Disposition").
+									Contains(Fmt(`{{disposition}}; filename="{{cid}}.{{format}}"`, row.Disposition, plainCID, row.Format)),
+							),
+					},
+					plain.RawData(),
+					Fmt("application/{{format}}", row.Format),
+				)...).
 			Append(
 				helpers.IncludeRandomRangeTests(t,
 					SugarTest{
@@ -323,7 +323,7 @@ func TestPlainCodec(t *testing.T) {
 				},
 			)
 
-		RunWithSpecs(t, *tests, specs.PathGatewayDAG)
+		RunWithSpecs(t, tests, specs.PathGatewayDAG)
 
 		if dagFormattedResponse != nil {
 			rangeTests := helpers.OnlyRandomRangeTests(t,
--- a/tests/path_gateway_dag_test.go
+++ b/tests/path_gateway_dag_test.go
@@ -253,8 +253,8 @@ func TestPlainCodec(t *testing.T) {
 						),
 				},
 				helpers.ByteRanges{
-					helpers.SimpleByteRange(0, 5, plain.RawData()[0:6]),
-					helpers.SimpleByteRange(10, 15, plain.RawData()[10:16]),
+					helpers.SimpleByteRange(0, 5),
+					helpers.SimpleByteRange(10, 15),
 				},
 				plain.RawData(),
 			)...).
@@ -277,8 +277,8 @@ func TestPlainCodec(t *testing.T) {
 							),
 					},
 					helpers.ByteRanges{
-						helpers.SimpleByteRange(0, 5, plain.RawData()[0:6]),
-						helpers.SimpleByteRange(10, 15, plain.RawData()[10:16]),
+						helpers.SimpleByteRange(0, 5),
+						helpers.SimpleByteRange(10, 15),
 					},
 					plain.RawData(),
 				)...).
@@ -303,8 +303,8 @@ func TestPlainCodec(t *testing.T) {
 							),
 					},
 					helpers.ByteRanges{
-						helpers.SimpleByteRange(0, 5, plain.RawData()[0:6]),
-						helpers.SimpleByteRange(10, 15, plain.RawData()[10:16]),
+						helpers.SimpleByteRange(0, 5),
+						helpers.SimpleByteRange(10, 15),
 					},
 					plain.RawData(),
 				)...).
@@ -360,8 +360,8 @@ func TestPlainCodec(t *testing.T) {
 						),
 				},
 				helpers.ByteRanges{
-					helpers.SimpleByteRange(0, 5, dagFormattedResponse[0:6]),
-					helpers.SimpleByteRange(10, 15, dagFormattedResponse[10:16]),
+					helpers.SimpleByteRange(0, 5),
+					helpers.SimpleByteRange(10, 15),
 				},
 				dagFormattedResponse,
 			)
@@ -644,8 +644,8 @@ func TestNativeDag(t *testing.T) {
 					),
 			},
 			helpers.ByteRanges{
-				helpers.SimpleByteRange(6, 16, dagTraversal.RawData()[6:17]),
-				helpers.SimpleByteRange(20, 25, dagTraversal.RawData()[20:26]),
+				helpers.SimpleByteRange(6, 16),
+				helpers.SimpleByteRange(20, 25),
 			},
 			dagTraversal.RawData(),
 		)...).Append(
@@ -663,8 +663,8 @@ func TestNativeDag(t *testing.T) {
 						),
 				},
 				helpers.ByteRanges{
-					helpers.SimpleByteRange(6, 16, dagTraversal.RawData()[6:17]),
-					helpers.SimpleByteRange(20, 25, dagTraversal.RawData()[20:26]),
+					helpers.SimpleByteRange(6, 16),
+					helpers.SimpleByteRange(20, 25),
 				}, dagTraversal.RawData(),
 			)...).Append(
 			helpers.RangeTestTransform(t,
@@ -681,8 +681,8 @@ func TestNativeDag(t *testing.T) {
 						),
 				},
 				helpers.ByteRanges{
-					helpers.SimpleByteRange(6, 16, dagTraversal.RawData()[6:17]),
-					helpers.SimpleByteRange(20, 25, dagTraversal.RawData()[20:26]),
+					helpers.SimpleByteRange(6, 16),
+					helpers.SimpleByteRange(20, 25),
 				},
 				dagTraversal.RawData(),
 			)...)
@@ -727,8 +727,8 @@ func TestNativeDag(t *testing.T) {
 				Response: Expect(),
 			},
 			helpers.ByteRanges{
-				helpers.SimpleByteRange(1, 5, dagJsonConvertedData[1:6]),
-				helpers.SimpleByteRange(93, 104, dagJsonConvertedData[93:105])},
+				helpers.SimpleByteRange(1, 5),
+				helpers.SimpleByteRange(93, 104)},
 			dagJsonConvertedData)
 
 		RunWithSpecs(t, rangeTests, specs.PathGatewayDAG)
@@ -767,8 +767,8 @@ func TestNativeDag(t *testing.T) {
 					),
 				Response: Expect(),
 			}, helpers.ByteRanges{
-				helpers.SimpleByteRange(1, 5, dagCborHTMLRendering[1:6]),
-				helpers.SimpleByteRange(93, 104, dagCborHTMLRendering[93:105])},
+				helpers.SimpleByteRange(1, 5),
+				helpers.SimpleByteRange(93, 104)},
 			dagCborHTMLRendering)
 
 		RunWithSpecs(t, rangeTests, specs.PathGatewayDAG)
--- a/tests/path_gateway_dag_test.go
+++ b/tests/path_gateway_dag_test.go
@@ -236,7 +236,7 @@ func TestPlainCodec(t *testing.T) {
 
 		tests := &SugarTests{}
 		tests.Append(
-			helpers.IncludeRangeTests(t,
+			helpers.IncludeRandomRangeTests(t,
 				SugarTest{
 					Name: Fmt(`GET {{name}} without Accept or format= has expected "{{format}}" Content-Type and body as-is`, row.Name, row.Format),
 					Hint: `
@@ -250,12 +250,11 @@ func TestPlainCodec(t *testing.T) {
 								Contains(Fmt(`{{disposition}}; filename="{{cid}}.{{format}}"`, row.Disposition, plainCID, row.Format)),
 						),
 				},
-				nil,
 				plain.RawData(),
 				Fmt("application/{{format}}", row.Format),
 			)...).
 			Append(
-				helpers.IncludeRangeTests(t,
+				helpers.IncludeRandomRangeTests(t,
 					SugarTest{
 						Name: Fmt("GET {{name}} with ?format= has expected {{format}} Content-Type and body as-is", row.Name, row.Format),
 						Hint: `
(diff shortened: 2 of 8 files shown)


tooling/test/validate.go
```diff
@@ -9,75 +10,92 @@ import (
...
+	checkOutput check.CheckOutput
+}
+
 func validateResponse(
 	t *testing.T,
 	expected ExpectBuilder,
 	res *http.Response,
-	localReport Reporter,
-) {
+) []testCheckOutput {
```
aschmahmann, line 26: We don't do the reporting here anymore so that it can be controlled by the caller (e.g. for `AnyOf` we don't fail the test unless all the options fail)

tooling/test/validate.go
```diff
@@ -9,75 +10,92 @@ import (
...
 	t.Helper()
 
+	var outputs []testCheckOutput
+
 	if expected.StatusCode_ != 0 {
+		output := testCheckOutput{testName: "Status Code", checkOutput: check.CheckOutput{Success: true}}
```
aschmahmann, line 32: It was a little awkward that StatusCode wasn't its own test, so it is now.

tooling/test/validate.go
```diff
@@ -9,75 +10,92 @@ import (
...
-			resBody, err := io.ReadAll(res.Body)
-			if err != nil {
-				localReport(t, err)
+		defer res.Body.Close()
+		resBody, err := io.ReadAll(res.Body)
+		if err != nil {
+			outputs = append(outputs, testCheckOutput{testName: "Body", checkOutput: check.CheckOutput{Success: false, Reason: err.Error()}})
+			return outputs
+		}
+		res.Body = io.NopCloser(bytes.NewBuffer(resBody))
```
aschmahmann, line 68: This was added so that we can reuse and query the responses

tooling/helpers/range.go
```diff
@@ -0,0 +1,176 @@
...
+
+	return rangeTest
+}
+
+// MultiRangeTestTransform takes a test where there is no "Range" header set in the request, or checks on the
+// StatusCode, Body, or Content-Range headers and verifies whether a valid response is given for the requested ranges.
+//
+// Note: HTTP Multi Range requests can be validly responded with one of the full data, the partial data from the first
+// range, or the partial data from all the requested ranges
+func MultiRangeTestTransform(t *testing.T, testWithoutRangeRequestHeader test.SugarTest, branges ByteRanges, fullData []byte) test.SugarTest {
```
aschmahmann, line 131: Note: We could have a combined function for Single + Multi ranges but seems like it'd require implementing `Clone` for Responses.

We can do `Clone` for responses if we want (I've got some code locally from an attempt), but the clones are not completely deep because we pass functions around for checks. Thoughts?

tooling/helpers/range.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return fmt.Sprintf("bytes=%s", strings.Join(rangeStrs, ","))
+}
+
+// SingleRangeTestTransform takes a test where there is no "Range" header set in the request, or checks on the
+// StatusCode, Body, or Content-Range headers and verifies whether a valid response is given for the requested range.
+//
+// Note: HTTP Range requests can be validly responded with either the full data, or the requested partial data
+func SingleRangeTestTransform(t *testing.T, baseTest test.SugarTest, brange ByteRange, fullData []byte) test.SugarTest {
+	modifiedRequest := baseTest.Request.Clone().Header("Range", brange.GetRangeString(t))
```
laurentsenta, line 98: Could you push that clone to the `Header` call? That would help push https://github.com/ipfs/gateway-conformance/issues/36 forward and hide the internals deeper.

tooling/test/sugar.go
```diff
@@ -208,6 +216,67 @@ func (e ExpectBuilder) BodyWithHint(hint string, body interface{}) ExpectBuilder
 	return e
 }
 
+func (e ExpectBuilder) Validate(t *testing.T, res *http.Response, localReport Reporter) {
+	t.Helper()
+
+	checks := validateResponse(t, e, res)
+	for _, c := range checks {
+		t.Run(c.testName, func(t *testing.T) {
```
laurentsenta, line 224: Awesome, that preserves the test structures 👍

tooling/test/sugar.go
```diff
@@ -208,6 +216,67 @@ func (e ExpectBuilder) BodyWithHint(hint string, body interface{}) ExpectBuilder
...
+func (e AllOfExpectBuilder) Validate(t *testing.T, res *http.Response, localReport Reporter) {
+	t.Helper()
+
+	for _, expect := range e.Expect_ {
+		expect.Validate(t, res, localReport)
+	}
+}
+
+type AnyOfExpectBuilder struct {
+	Expect_ []ExpectBuilder `json:"expect,omitempty"`
```
laurentsenta, line 251: Could we use a `ExpectValidator` here?

tooling/test/sugar.go
```diff
@@ -208,6 +216,67 @@ func (e ExpectBuilder) BodyWithHint(hint string, body interface{}) ExpectBuilder
...
+
+func AllOf(expect ...ExpectValidator) AllOfExpectBuilder {
+	return AllOfExpectBuilder{Expect_: expect}
+}
+
+func (e AllOfExpectBuilder) Validate(t *testing.T, res *http.Response, localReport Reporter) {
+	t.Helper()
+
+	for _, expect := range e.Expect_ {
+		expect.Validate(t, res, localReport)
```
laurentsenta, line 246: What about we wrap this in

```go
t.Run(fmt.Sprintf("Check %d", i), {
    ...
})
```
?

That'll give us a clearer structure for reporting.

tooling/test/sugar.go
```diff
@@ -208,6 +216,67 @@ func (e ExpectBuilder) BodyWithHint(hint string, body interface{}) ExpectBuilder
...
+}
+
+var _ ExpectValidator = (*AnyOfExpectBuilder)(nil)
+
+func AnyOf(expect ...ExpectBuilder) AnyOfExpectBuilder {
+	return AnyOfExpectBuilder{Expect_: expect}
+}
+
+func (e AnyOfExpectBuilder) Validate(t *testing.T, res *http.Response, localReport Reporter) {
+	t.Helper()
```
laurentsenta, line 261: We put a lot of emphasis on reporting for this test suite: one of our goals is to have [really clear dashboards](https://singulargarden.github.io/gateway-conformance/tests/testtrustlesscarpathing--get-default-car-response-for-non-existing-file--accept-header-/) that maintainers (and potential users) can check to compare implementations and expectation.

In the case of `AnyOf`, we should make sure that reports are identical between gateways and that the supported features are explicit (range requests support vs naive responses, etc). 

Naively, that function would look something like:

```go
originalFailure := localReport.inCaseOfFailure // assume t.Fail but that might change with recursion
hasSucceeded := false

t.Run(fmt.Sprintf("Check %d", i), {
   if (hasSucceeded || !isLastCheck) {
      localReport.inCaseOfFailure = t.Skip
   } else {
      localReport.inCaseOfFailure = originalFailure
   }

   hasSucceeded = hasSucceeded || expect.Validate(t, res, localReport)    
})

localReport.inCaseOfFailure = originalFailure
return hasSucceeded
```

tooling/helpers/range.go
laurentsenta: You thought about this more than I did, why do we need Clone?
I'd assume the Single case is "just" the Many case with N = 1.

Worst case scenario, I guess we could even do something like this:

```go
func WithRangesCheck(t *testing.T, test test.SugarTest, branges ByteRanges, fullData []byte) {
    if (branges.length == 0) {
      panic()
    }
    else if (branges.length == 1) {
       return singleRangeTransform(...)
    }
    else {
       return multipleRangeTransform(...)
    }
}

// return a ByteRangeS to please type checker AND users
func SimpleByteRange(from, to uint64, data []byte) ByteRanges {
   ...
}

func ManyByteRanges(...ByteRanges) {
   ...
}
```

Thinking about usage & consistency sugar API we could even do:

```go
ByteRange(1, 2, bytes)
  .And(5, 6, bytes)
  .And(8, 12, bytes)
```

Regarding Clone:
the `CheckFunc` is more of a temporary bandaid for cases that are not yet supported by the Sugar API. I'd avoid designing around it. I'd implement the `Clone()` function as a deep clone and assume we'll never need shallow cloning. It's easier to relax constraints later than add them. We could provide a `CheckPureFunc` for internal checks, and _trustworthy_ users

tests/path_gateway_dag_test.go
```diff
@@ -563,10 +669,168 @@ func TestNativeDag(t *testing.T) {
...
+					Path("/ipfs/{{cid}}/", dagCborCID).
+					Headers(
+						Header("Accept", "text/html"),
+						Header("Range", "bytes=1-5,93-104"),
+					),
+				Response: Expect(),
+			}, helpers.ByteRanges{
+				helpers.SimpleByteRange(1, 5, dagCborHTMLRendering[1:6]),
+				helpers.SimpleByteRange(93, 104, dagCborHTMLRendering[93:105])},
+				dagCborHTMLRendering),
```
laurentsenta, line 829: Each of these test repeat the range parameters,

```go
...

						Header("Range", "bytes=1-5,93-104"),

...

				helpers.SimpleByteRange(1, 5, dagCborHTMLRendering[1:6]),
				helpers.SimpleByteRange(93, 104, dagCborHTMLRendering[93:105])},

...
```

Would that make sense to either, be explicit about generation:

```go
						Header("Range", "bytes=GENERATED"),
```

check that the ranges match?

tooling/helpers/range.go
aschmahmann: i.e. to make the call to `Header` automatically perform the clone? I think that's probably best done in a follow-up PR since the same could probably be said for basically everywhere `Clone` is currently used.

tooling/test/sugar.go
aschmahmann: Ah ok, so we want the same number of testcases when possible. Can do.

tooling/helpers/range.go
aschmahmann: > Worst case scenario, I guess we could even do something like this:

What I was thinking was that we could make a wrapper where the user passes the test for a standard non-range request and we could return 2 tests `(singleRange, multiRange)` (or maybe 3 if we include `noRange`) and so make it easier to test for ranges basically anywhere we wanted to.

tests/path_gateway_dag_test.go
```diff
@@ -563,10 +669,168 @@ func TestNativeDag(t *testing.T) {
...
+				if innerCheck.Success {
+					dagJsonConvertedData = t
+					return true
+				}
+				return false
+			})),
+		},
+	}, specs.PathGatewayDAG)
+
+	if dagJsonConvertedData != nil {
```
aschmahmann, line 750: @laurentsenta if there's an error here and this ends up being nil the number of tests might not match exactly between runs against different gateways. I copied this pattern from the prior range tests. Is this ok?

tests/path_gateway_dag_test.go
aschmahmann: the passed header is in the request, the SimpleByteRange is for the response. Should be able to wrap them all up together (i.e. insist the request doesn't have the Range header) and auto-generate the header in the request using `request.Clone` though.

tooling/helpers/range.go
laurentsenta: Sounds good!

tests/path_gateway_dag_test.go
laurentsenta: 🤦  my bad, I got confused between the generated part and the request, that way looks good.

tests/path_gateway_dag_test.go
laurentsenta: To me that's acceptable: we should make sure the tooling / framework maintains the counts,
That code is user-side, and it's a bandaid until we solve https://github.com/ipfs/gateway-conformance/issues/98, the simpler we keep it, the easier to replace it.

tooling/test/sugar.go
aschmahmann: It's doable, but not without some changes because the `AnyOfExpectBuilder`'s `Validate` function uses `validateResponse` which operates on an `ExpectBuilder` and returns the checks so they can be expected.

Up for proposals on changing the `Validate` function (or bundling everything into the `ExpectBuilder` struct) to support `AnyOf` working on multiple layers (i.e. on top of `AnyOf` or `AllOf`), although might also be ok to wait until we need it. WDYT?

tooling/test/test.go
```diff
@@ -14,12 +14,17 @@ type SugarTest struct {
...
+	Response  ExpectValidator
 	Responses ExpectsBuilder
 }
 
 type SugarTests []SugarTest
 
+func (s *SugarTests) Append(tests ...SugarTest) *SugarTests {
+	*s = append(*s, tests...)
+	return s
+}
```
aschmahmann, line 26: Added this helper since the pattern with struct literals we were using didn't work with helpers that returned multiple tests.

WDYT?

tooling/test/sugar.go
```diff
@@ -208,6 +218,148 @@ func (e ExpectBuilder) BodyWithHint(hint string, body interface{}) ExpectBuilder
...
+
+		t.Run(fmt.Sprintf("Check %d", i),
+			func(t *testing.T) {
+				if !responseSucceeded {
+					for _, c := range checks {
+						if c.checkOutput.Success {
+							t.Logf("Test %s passed", c.testName)
+						} else {
+							t.Logf("Test %s failed with: %s", c.testName, c.checkOutput.Reason)
+						}
```
aschmahmann, line 340: @laurentsenta this can be pretty noisy, I'm not sure if this is what you were looking for. Your suggestion https://github.com/ipfs/gateway-conformance/pull/162/files#r1322581089 implied a bunch of plumbing under the hood around `localReport` which I wasn't sure made sense to bundle in here as I'd like to get the basics here closed out.

If you think it's needed say so though and @hacdias or I can take a look

tooling/helpers/range.go
```diff
@@ -0,0 +1,247 @@
...
+		dataLen := len(fullData)
+		if dataLen < 10 {
+			panic("transformation not defined for data smaller than 10 bytes")
+		}
+
+		branges = ByteRanges{
+			SimpleByteRange(7, 9),
+			SimpleByteRange(1, 3),
+		}
+	}
```
aschmahmann, line 225: Having some flexibility here seemed useful, I was finding that coming up with these numbers when they were in the tests was mostly a repeated exercise around what numbers wouldn't panic and what might be some wacky orders to try.

We could make this randomized, use the data size to determine the ranges, try things like suffix requests, etc. This is mostly for illustration and was a trivial example. Suggestions welcome.

tooling/test/sugar.go
```diff
@@ -227,6 +238,142 @@ func (e ExpectBuilder) BodyWithHint(hint string, body interface{}) ExpectBuilder
...
+				responseSucceeded = false
+				break
+			}
+		}
+		if responseSucceeded {
+			hadASuccessfulResponse = true
+		}
+
+		t.Run(fmt.Sprintf("Check %d", i),
+			func(t *testing.T) {
```
aschmahmann, line 347: @laurentsenta could you help fill in how you'd like the spec logging to work for the `AnyOf` and `AllOf` builders? While I think I did this correctly for the existing `ExpectBuilder` I'm not sure how these are supposed to work given I'm not sure what the requirements on the spec metadata emitting are supposed to look like.

tooling/test/sugar.go
laurentsenta: Sure, let's hold until we need it 👍

tooling/test/test.go
laurentsenta: SGTM, anything stops us from using a Value Received instead of a pointer here?

That'll make the test code more consistent and remove a side-effect.

tooling/test/sugar.go
laurentsenta: Agreed, this is more complex than it needs to, afaii you'd need the recursivity on `Validate()`, and the side-effect on local report to make this a few loc. Let's keep it simple: feel free to drop that part, as long as the public API works for you, we'll revisit this later.

tooling/test/sugar.go
laurentsenta: We're defining our tests structure with data instead of go code, but as long as we nest `t.Run` calls in a way that map the test hierarchy, the `validateResponse` and `runRequest` should take care of logging the specs metadata (and hints and all) to the right location.

For example, we define this test structure with a piece of `TestSugar{}`: 

```yaml
TESTS:
  test1:
    response1:
      - check1
      - check2
  test2:
    AND:
      - response1:
          - check1
          - check2
      - response2:
          - check3
          - check4
  test3:
    OR:
      - response1:
          - check1
          - check2
      - AND:
          - response2:
              - check3
              - check4
          - response3:
              - check5
              - check6
```

We want to run something like:

```go
t.Run("tests",
  ...
  t.Run("test1",
    ...
    t.Run("response1",
      ...
      t.Run("check1",
        ...
      )
      t.Run("check2",
        ...
      )
    )
  )
  t.Run("test2",
    ...
    t.Run("AND",
      ...
      t.Run("response1",
        ...
      )
      t.Run("response2",
        ...
      )
    )
  )
  ... etc
)
```

See my comment below, we can keep it simple, I'll revisit this later.

tooling/helpers/range.go
laurentsenta: Long answer on purpose, I'll probably reuse this later:

The idea behind this test suite is to have a test suite that resembles the specs, and that can be read, understood, and written by someone who is not a maintainer or even a go programmer. If you can read English, you can understand the specs and the test definitions.

When we move test-related data out of the test definition (or we add memory management tidbits see my comment below), we change the nature of the project.

To me, the architecture of the conformance test should be:

- ./tests => targets humans, maps to specs, self-contain, written with English-like sugar
- ./tooling/helpers => targets devs, data transformation operations, based on the lower-APIs, written with some go knowledge. Expose English-API to humans reading `./tests`
- ./tooling => targets go maintainers, yolo

That split is one of the reasons we didn't "just" translate shell into go code but went for a data-driven approach.

That code seems to move test data out of the test definition for the sake of making it generalizable and saving lines of go code. Could we define this as a const in the test file itself and reuse it?

laurentsenta (approved):
The code looks mostly fine, feel free to take into account feedback or merge as is.

Could you add a line in the Changelog?

lidel (dismissed):
I am not sure if we should merge this without cleaning up DSL first.
We may not have time to clean up later. See comments inline.

tooling/test/test.go
laurentsenta: @hacdias do you have more details? Else I'd rather keep a Value here

tests/path_gateway_dag_test.go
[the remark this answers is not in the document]
laurentsenta: 👎  on Idea B, we have other options that don't require growing the SugarTest structure,

👍  Option A, it seems to be @aschmahmann's initial sugar without the custom types,
I liked that it solved one thing: "I want to take a request/response and make it ipfs-spec aware with multiple valid responses".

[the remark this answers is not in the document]
lidel: Updated (A) to include `IncludeRandomRangeTests` (will pick 1,2 ranges within body size)  in addition to `IncludeRangeTests` which takes explicit ranges.

I think if we have that, we could live with the resulting DSL (the reader can infer what is happening without having to read helper code).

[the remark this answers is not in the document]
hacdias: @lidel @laurentsenta @aschmahmann I refactored with @lidel's suggestion A.

tests/path_gateway_dag_test.go
```diff
@@ -231,86 +232,121 @@ func TestPlainCodec(t *testing.T) {
 		plainCID := plain.Cid()
 		plainOrDagCID := plainOrDag.Cid()
 
-		tests := SugarTests{
-			{
-				Name: Fmt(`GET {{name}} without Accept or format= has expected "{{format}}" Content-Type and body as-is`, row.Name, row.Format),
-				Hint: `
+		var dagFormattedResponse []byte
+
+		tests := &SugarTests{}
```
laurentsenta, line 237: Could you stick to values?

(details in https://github.com/ipfs/gateway-conformance/pull/162/files#r1333946686)

laurentsenta (changes requested):
I left a comment regarding values / pointer,
could you also add a line to the Changelog?

Everything else lgtm, thanks 🙏

hacdias:
@laurentsenta I changed it to use values!

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added tests for HTTP Range requests, as well as some basic helpers for `AnyOf` and `AllOf`. [PR](https://github.com/ipfs/gateway-conformance/pull/162)
 ## [0.3.1] - 2023-09-15
```
laurentsenta, line 12:
```suggestion
## Unreleased
### Added
- Added tests for HTTP Range requests, as well as some basic helpers for `AnyOf` and `AllOf`. [PR](https://github.com/ipfs/gateway-conformance/pull/162)

## [0.3.1] - 2023-09-15
```

laurentsenta (approved):
lgtm!

PR #173, opened 2023-10-02
Tests for `/ipns` paths and `Cache-Control` HTTP header

hacdias:
This updates the conformance tests to bring them in line with the changes on the following PRs:

- https://github.com/ipfs/specs/pull/436
- https://github.com/ipfs/boxo/pull/459
- https://github.com/ipfs/kubo/pull/10115

Currently, we are not allowing `Cache-Control` headers in `/ipns` paths, but that's about to change.

### DIFF
--- a/.github/workflows/test-kubo-e2e.yml
+++ b/.github/workflows/test-kubo-e2e.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: ['latest', 'master', 'ns-cleanup']
+        target: ['latest', 'master']
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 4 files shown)


hacdias:
@lidel since the specifications say _SHOULD_ all the time regarding `Cache-Control` and `Last-Modified`, I find it hard to write meaningful tests. For `Cache-Control`, I changed from "must not have" to "either is empty, or is `public, max-age=*`". Regarding `Last-Modified`, I simply did not add checks. The spec says it should be the resolution time. So that means it can either be set or not. The only thing I could do is checking if the date is valid if present, but that seems very meta into HTTP.

lidel (approved):

lidel:
Ack, I think this is sensible for now. We will refine once DNSLink  side of things is tackled in the future PRs.

hacdias:
As this needs to be merged, and **published** before merging the Boxo and Kubo PRs, I am going ahead and merge this once CI completes (just removed the Boxo branch).

PR #177, opened 2023-10-26
feat: fleek deploy

laurentsenta:


.github/workflows/deploy-pages.yml
lidel, line 1: @laurentsenta this works on CI, but there is no way to build locally, or on Fleek.
Was this intentional? Makes it very hard to maintain the thing, or move between providers. 

I was unable to build locally cause there was no obvious script to run. 
I've pushed my changes into this PR and then manually run Deploy in fleek app which builds from `feat/fleek`, but nothing changed. https://conformance-ipfs.on.fleek.co still shows old website. 

Lmk if i am doing something wrong – meta comment is to simplify this.

Would it be possible to move build steps from this YAML into `Makefile` target named `www`?
If we could build the website locally with `make www`, then we could do the same on Fleek, removing the limitation of having website build being possible only in CI.

(Alternative ideas welcome, not being able to build locally is something we need to fix)

.github/workflows/deploy-pages.yml
```diff
@@ -69,6 +70,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       HUGO_VERSION: 0.117.0
+    permissions:
+      contents: write
```
lidel, line 74: Why we need this + `git commit` below?

lidel (changes requested):

.github/workflows/deploy-pages.yml
laurentsenta: @lidel (see the other comment) we deploy the `fleek-deploy` branch to fleek.
We build within Github's CI, then commit & push to the branch used by fleek.

.github/workflows/deploy-pages.yml
laurentsenta: @lidel We deploy `fleek-deploy`. `feat/fleek` is a temporary feature branch; it's used to trigger the build until we merge to main.

We're not "just" building a static website from the local content; At every build, we get the latest artifacts & assets from other repositories. So rational here was that Github tokens & Github-related code would stay in Github build system. We'd "just" deploy the last step (`fleek-deploy`) to fleek.

I get it's not super obvious how the whole thing work, I'll move the code to Makefile and scripts 👍

laurentsenta: @lidel I simplified the workflow so you can run it locally using `make website` (documented in the README).

Regarding building on fleek directly, do you guys have a way to manage IPFS' Github tokens in fleek right now? (tokens expire at best after 12 months).

With changes in the org, I'd recommend sticking to a workflow that don't require maintenance or moving secrets around:

- we build on Github (now using the same `make website` command you'd use),
- then we push the result to the branch [`fleek-deploy`](https://github.com/ipfs/gateway-conformance/tree/fleek-deploy),
- fleek deploys that branch.

www/static/logo.png
galargh, line 1: I was wondering if we should change the logo since we adopted the 🐙 as, at least interim, logo of IPDX.

galargh (approved):

.github/workflows/deploy-pages.yml
```diff
@@ -81,55 +39,18 @@ jobs:
...
-    environment:
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
-    runs-on: ubuntu-latest
-    needs: build
-    steps:
-      - name: Deploy to GitHub Pages
-        id: deployment
-        uses: actions/deploy-pages@v1
+            ./artifacts
```
galargh, line 56: I find this helpful for dealing with eof newlines.

https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files#:~:text=Additionally%2C%20in%20the%20Settings%20page,there%20isn't%20already%20one.

Issue #184, opened 2023-12-04
docs: should add instructions to run from source


SgtPooki:
What I did

```bash
make gateway-conformance
go run ./cmd/gateway-conformance/main.go test # ...args
```

If there's a better way please let me know.

Also of note.. you cannot `go install` this package, but it would make testing much easier..

```
╰─ ✘ 1 ❯ go version
go version go1.21.4 darwin/arm64

╰─ ✘ 1 ❯ go install github.com/ipfs/gateway-conformance@latest
go: downloading github.com/ipfs/gateway-conformance v0.4.2
go: github.com/ipfs/gateway-conformance@latest: module github.com/ipfs/gateway-conformance@latest found (v0.4.2), but does not contain package github.com/ipfs/gateway-conformance
```

SgtPooki:
`go run github.com/SgtPooki/gateway-conformance@1f9d1f393d572b863de9b6303cd551573ac77a4c` 

:shrug:

laurentsenta:
@SgtPooki, thanks for opening this, could you try going through the documentation and details which parts are broken?

https://github.com/ipfs/gateway-conformance/blob/main/docs/development.md

SgtPooki:
Some errors when following instructions at https://github.com/ipfs/gateway-conformance/blob/main/docs/development.md

```
╰─ ✔ ❯ make provision-kubo
find ./fixtures -name '*.car' -exec ipfs dag import --stats --pin-roots=false {} \;
Imported 3 blocks (119 bytes)
Imported 9 blocks (1541 bytes)
Imported 2 blocks (128 bytes)
Imported 11 blocks (759 bytes)
Imported 9 blocks (1462 bytes)
Imported 10 blocks (1538 bytes)
Imported 9 blocks (1541 bytes)
Imported 4 blocks (208 bytes)
Imported 243 blocks (74982 bytes)
Imported 3 blocks (2215 bytes)
Imported 10 blocks (623 bytes)
Imported 5 blocks (224 bytes)
Imported 1 blocks (26 bytes)
Imported 1 blocks (44 bytes)
Imported 3 blocks (148 bytes)
Imported 4 blocks (184 bytes)
Imported 11 blocks (711 bytes)
Imported 1 blocks (52 bytes)
Imported 3 blocks (231 bytes)
Imported 1 blocks (17 bytes)
Imported 3 blocks (139 bytes)
Imported 4 blocks (176 bytes)
Imported 10 blocks (623 bytes)
[...]

PR #192, opened 2023-12-11
fix(make): correctly infer ipns name from record filename

hacdias:
Fixes the issue noted by @SgtPooki (#184) regarding `make provision-kubo`. It seems we forgot to update the Makefile to ignore everything after the `_` when we starting adding more information to the record filename.


SgtPooki (approved):
can confirm this fixes things for me when i run `make provision-kubo`

PR #202, opened 2024-04-11
feat: remove range test for dag-cbor as text/html

hacdias:
As per our conversation [here](https://github.com/ipfs/boxo/pull/587#pullrequestreview-1992168667), this removes the range test for `dag-cbor` objects that have been requested as `text/html`.

### DIFF
--- a/tests/path_gateway_dag_test.go
+++ b/tests/path_gateway_dag_test.go
@@ -694,7 +694,6 @@ func TestNativeDag(t *testing.T) {
 		RunWithSpecs(t, rangeTests, specs.PathGatewayDAG)
 	}
 
-	var dagCborHTMLRendering []byte
 	RunWithSpecs(t, SugarTests{
 		SugarTest{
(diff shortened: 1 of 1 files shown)


tests/path_gateway_dag_test.go
```diff
@@ -704,34 +703,9 @@ func TestNativeDag(t *testing.T) {
...
 				),
-			Response: Expect().Body(Checks("", func(t []byte) bool {
-				innerCheck := Contains("</html>").Check(string(t))
-				if innerCheck.Success {
-					dagCborHTMLRendering = t
-					return true
-				}
-				return false
-			})),
+			Response: Expect().Body(Contains("</html>")),
```
lidel, line 706: :thought_balloon:  Provides forcing function for implementations to provide meaningful HTMP preview for CBOR in browsers, but without forcing any specific UI or behavior.

lidel (approved):

PR #204, opened 2024-04-11
release: 0.5.1

lidel:
Releasing as patch release because removal of the test is backward-compatible.


lidel:
@hacdias sorry, I see you started doing this, I'm closing this one, will let you do it

PR #205, opened 2024-04-17
fix: relax dag-cbor error check

lidel:
Thanks for noticing this @SgtPooki, agree we should not care about body, we care about status code.


SgtPooki (approved):

hacdias (approved):

hacdias (changes requested):
Please add a line in the changelog before merging. Maybe we should add a CI check for that. Otherwise we'll have to go back and write the changelog when we want to release. Other than that, lgtm :)

PR #206, opened 2024-05-01
docs: add install instructions for cli use

SgtPooki:
Fixes https://github.com/ipfs/gateway-conformance/issues/184

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ Two high level [commands](/docs/commands.md) exist:
 ### CLI
 
 ```console
+$ # Install the gateway-conformance binary
+$ go install github.com/ipfs/gateway-conformance/cmd/gateway-conformance@latest
 $ # run subdomain-gateway tests against endpoint at http://localhost:8080 output as JSON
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,8 @@ Two high level [commands](/docs/commands.md) exist:
 ### CLI
 
 ```console
+$ # Install the gateway-conformance binary
+$ go install github.com/ipfs/gateway-conformance/cmd/gateway-conformance@latest
```
SgtPooki, line 69: For others like me, it might be useful to point to the `go help install` docs that mention where the binary will be installed:

> Executables are installed in the directory named by the GOBIN environment
variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH
environment variable is not set. Executables in $GOROOT
are installed in $GOROOT/bin or $GOTOOLDIR instead of $GOBIN.

SgtPooki: it might be useful to mention adding `$GOPATH/bin` or `$HOME/go/bin` to PATH if the user is not already familiar with the Go ecosystem.

PR #207, opened 2024-05-16
fix: Header().Has works properly for checking multiple values

SgtPooki:
Initial commit shows the problem mentioned in https://github.com/ipfs/gateway-conformance/issues/186

later commits fix the failing test

fixes https://github.com/ipfs/gateway-conformance/issues/186
contributes to #21

PR #208, opened 2024-05-20
chore: release 0.5.2

lidel:


lidel:
Ugh, we have "green" CI for "e2e kubo latest|main",  but Kubo is failing 9 conformance tests, and only indicator is a number in a comment that people  are trained (thanks to codecov) to ignore. 

I think the problem was introduced in #207, Kubo fails because Kubo sends header more than once. 


## TODO

- [ ] make kubo e2e fail CI as RED, if any failures
- [ ] make kubo e2e pass

lidel:
Added normalization in https://github.com/ipfs/gateway-conformance/pull/208/commits/ca15cc851017a8f093be0b2452fbded362f107f2, it saves us headache of  having two code paths in every test sugar that has to operate on header values.

PR #211, opened 2024-05-24
feat: improved subdomain-url handling

### LINKED ISSUE #185: bug: gateway-conformance subdomains are inflexible and setup is not clear
I'm trying to improve on conformance of [helia-http-gateway](https://github.com/ipfs/helia-http-gateway/pull/57), and I am continuously getting failures from the gateway-conformance tool (via docker and directly with `go run ./cmd/gateway-conformance/main.go test # ...`) 

I believe these may be related to whatever proxying thing is happening under the hood, because i've seen this error more than a few times:

```
Error: Querying http://localhost/ipfs/bafkreicysg23kiwv34eg2d7qweipxwosdo2py4ldv42nbauguluen5v6am/ failed: Get "http://localhost/ipfs/bafkreicysg23kiwv34eg2d7qweipxwosdo2py4ldv42nbauguluen5v6am/": proxyconnect tcp: dial tcp [::1]:8090: connect: connection refused
```

Some questions:

1. If I am specifying `--gateway-url 'http://localhost:8090' --subdomain-url 'http://localhost:8090'`, why is the gateway-conformance tool querying `localhost` with no port?
2. Why are dials going to an ip6 address (that is not supported on my machine) and can I disable that?
3. Can we disable the proxy entirely? is there a way to run tests directly against a gateway and verify the actual HTTP headers and responses?

lidel:
This PR implements cleanup described in https://github.com/ipfs/gateway-conformance/issues/185#issuecomment-2128307963 and close #185 for good.

- [x] removed implicit tests based on `SubdomainLocalhostGatewayURL = http://localhost`
  - [x] subdomain tests are now run only against origins defined via `--subdomain-url`
  - [x] `--subdomain-url` defaults to `http://example.com:8080` now
- [x] DNSLink gateway tests are no longer tied to origin passed via `--subdomain-url`
  - [ ] turn "subdomain dnslink" fixtures into "domain" ones
- [x] cleanup `tooling/helpers/subdomain.go` – mangling of `Host`, `URL` and `Path` happens there and in `./tooling/test/run.go`, we should refactor tests to set `Path`  + optional `Host` header`; proxy tests should be in separate file that can be disabled 
- [x] produce dnslink fixtures in `IPFS_NS_MAP` format
- [x] e2e kubo CI reports no Failures


simulified:
sup
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 5; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

lidel:
A picture for future reference, in case the exploit from https://github.com/ipfs/gateway-conformance/pull/211#issuecomment-2155691144 (comment above) gets deleted ;-)

PR #218, opened 2024-07-26
fix: Verify Cache-Control header for UnixFS directories

acejam:
This PR relies on and is related to: https://github.com/ipfs/boxo/pull/639


lidel:
Thanks! I was not able to push small refactor, so will continue in https://github.com/ipfs/gateway-conformance/pull/219

PR #219, opened 2024-07-29
fix: Cache-Control for generated UnixFS dir listings on /ipfs

lidel:
Applies https://github.com/ipfs/gateway-conformance/pull/218 in a backward-compatible way, allowing us to ship as a patch release while we finalize the specs.


lidel:
Tested in https://github.com/ipfs/kubo/pull/10464 and looked green threre, we cna ship this as v0.6.1 to unblock https://github.com/ipfs/boxo/pull/639

Issue #164, opened 2023-09-15
Implement a nicer status range check


laurentsenta:
@hacdias implemented `StatusBetween(500, 599)`, it's easy to implement and does the job 👍 

For the sake of readability & user-friendliness, we could implement a _nicer_ matcher like `.Status("5xx")`.

see https://github.com/ipfs/gateway-conformance/pull/157#discussion_r1319568532

guillaumemichel:
Triage notes:
* `StatusBetween(500, 599)` should be good enough
* It is better to be explicit rather than checking for `"5xx"`

PR #220, opened 2024-08-01
Add: StatusMatch function for elegantly handingling Status Checks

abhiraj-mengade:
This PR adds a function `StatusMatch` to handle status checks for ranges of HTTP status codes.

Aims to resolve the issue #164

### DIFF
--- a/tooling/test/sugar.go
+++ b/tooling/test/sugar.go
@@ -4,6 +4,9 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/ipfs/gateway-conformance/tooling"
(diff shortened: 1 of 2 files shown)


tooling/test/sugar.go
```diff
@@ -162,6 +165,34 @@ func (e ExpectBuilder) Status(statusCode int) ExpectBuilder {
...
+	if err != nil {
+		panic(fmt.Sprintf("invalid status pattern: %v", err))
+	}
+
+	e.StatusCodeFrom_ = from
+	e.StatusCodeTo_ = to
+	return e
+}
+
 func (e ExpectBuilder) StatusBetween(from, to int) ExpectBuilder {
```
SgtPooki, line 196: Are we using this function elsewhere? @lidel Are there any cases where we would want to do `501-503` or other?

SgtPooki (changes requested):
We should also add some tests for this function to ensure it works as expected.

tooling/test/sugar.go
lidel: I don't think so, in practice we care about specific code, or 4XX or 5XX

PR #221, opened 2024-08-09
fix(tar): test status and not body

gammazero:
The error message in boxo was [fixed](https://github.com/ipfs/boxo/pull/653/files#diff-1f636927091be5710efa23d3ac59429fe81b9ddcc290ce4e6b276539a665f167L13-R13) to say "not allowed" instead of "now allowed". The message that the gateway conformance test looks for needs to match.


lidel (approved):
thanks! refactored so this can be released as a v6 patch

(fleek error is not related to this PR - we have a website which no longer builds correctly, but that does not impact test suite, will fix in spare time in future PRs)

PR #252, opened 2025-11-26
test!(ipip-523/v0.9): prefer format query param over accept header

achingbrain:
- https://github.com/ipfs/specs/pull/523

When both are present, allow the query param to override the accept header.

BREAKING CHANGE: these tests expect URL query parameters to take precedence over Accept header, which reverses the previous behavior, needs to be released as minor release.

### DIFF
--- a/.github/workflows/test-kubo-e2e.yml
+++ b/.github/workflows/test-kubo-e2e.yml
@@ -31,10 +31,18 @@ jobs:
         uses: ./gateway-conformance/.github/actions/extract-fixtures
         with:
           output: fixtures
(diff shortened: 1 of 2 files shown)


lidel (approved):
Thank you, I've added a few more tests and implemented the IPIP-523 in:

- https://github.com/ipfs/boxo/pull/1074
- https://github.com/ipfs/kubo/pull/11086  (this is where we use everything end-to-end and want CI to be green – and it is!)

and works as expected.  

If no concerns I'm going to release this later this week after I finish reviewing other IPIPs.

